PR #4, opened 2016-10-06
fixes for building with GCC 6 under Linux

jcelerier:


src/ConnectionGeometry.cpp
```diff
@@ -1,4 +1,5 @@
 #include "ConnectionGeometry.hpp"
+#include <cmath>
```
jcelerier: IIRc it's std::sin, std::tan or std::abs

paceholder: Wrong. Sorry. It is std::abs()

jcelerier: right, that was the one  !

PR #8, opened 2016-11-21
make DataModelRegistry non-static

russelltg:
DataModelRegistry shouldn't be static, often the application will want to swap registries.


src/DataModelRegistry.hpp
```diff
@@ -56,10 +56,18 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
   using RegisteredModelsMap =
           std::unordered_map<QString, RegistryItemPtr>;
 
+  DataModelRegistry() = default;
```
paceholder: For completeness:
`~DataModelRegistry() = default;`   ?

russelltg: Done!

examples/calculator/main.cpp
```diff
@@ -15,27 +15,26 @@
 #include "MultiplicationModel.hpp"
 #include "DivisionModel.hpp"
 
-static bool
+static std::unique_ptr<DataModelRegistry>
```
paceholder: I made some comments from my phone but can't see them now.

What is the benefit of using `unique_ptr` ? 
I would return just `DataModelRegistry`

jcelerier: +1

src/DataModelRegistry.hpp
```diff
@@ -56,10 +56,19 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
...
+  ~DataModelRegistry() = default;
+
+  DataModelRegistry(const DataModelRegistry&) = delete;
+  DataModelRegistry(DataModelRegistry&&) = default;
+
+  DataModelRegistry& operator=(const DataModelRegistry&) = delete;
```
paceholder: Codestyle: `const` follows the type everywhere in the code

russelltg:
Aah thanks, unique ptr isn't needed at all, your right. No polymorphism. I'll patch this PR up when I get out of bed :).

src/DataModelRegistry.hpp
paceholder: It is a minor code style thing:
instead of 
`function_name(const Type &)`
I write everywhere
`function_name(Type const &)`

russelltg:
It could be true that the best approach is instead of lvalue `DataModelRegistry`, a `shared_ptr<DataModelRegistry>` is the way to go so multiple `FlowScene`s can share a `DataModelRegistry`

src/DataModelRegistry.hpp
russelltg: Fixed!

russelltg:
I'll fix the merge errors

paceholder:
Please, squash commits after fixing the conflicts.

russelltg:
Wouldnt we loose all the PR history then?

paceholder:
I do now know how github works. For example Atlassian's Bitbucket allows you to remove old commits and push new (squashed\modified), you are still able to see removed commits and the comments history

russelltg:
I'll try it, we'll see!

russelltg:
Weird there shouldn't be merge errors

paceholder:
Sorry I did a mess in the master. I will let you know when it is fixed.

paceholder:
I tried to move to C++14 and unexpectedly had some troubles with continuous integration service (travis).

Your changes were merged manually and are currently in the master.

PR #11, opened 2016-11-22
Use std::shared_ptr for storing registries so editors can share them

russelltg:
Use case:
You are writing an editor for graphs (like I am) and you want to have multiple graphs that share the same node types. These node types could change throughout the program execution, it would be a pain in the butt to copy them all for each update to the registry. Therefore, a shared registry is the way to go for this scenario (unless you have any other ideas)

### DIFF
--- a/src/FlowScene.hpp
+++ b/src/FlowScene.hpp
@@ -23,8 +23,7 @@ class NODE_EDITOR_PUBLIC FlowScene
 {
 public:
 
-  // Use && to assure that a rvalue reference is passed and the caller has to use std::move() or an rvalue
-  FlowScene(DataModelRegistry && registry = {});
+  FlowScene(std::shared_ptr<DataModelRegistry> registry = std::make_shared<DataModelRegistry>());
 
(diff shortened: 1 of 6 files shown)


paceholder:
Could you please make one commit? And happy Thanksgiving :)

russelltg:
Okay everything should be cleaned up now.

Happy Thanksgiving!

src/FlowScene.hpp
```diff
@@ -60,15 +59,13 @@ class NODE_EDITOR_PUBLIC FlowScene
...
+  registry() {
+    return *_registry;
   }
 
   void
-  setRegistry(DataModelRegistry && registry)
-  {
-    _registry = std::move(registry);
+  setRegistry(std::shared_ptr<DataModelRegistry> registry) {
+    if(registry) _registry = std::move(registry); // make sure _registry cannot be null
```
paceholder: It is fine to move in a null `std::shared_ptr`

http://cpp.sh/7qm5l

PR #16, opened 2016-11-28
Add headers in include that should be there

russelltg:
These are clearly meant to be exposed because there are functions returning pointers to these objects, but..there is no way to include them other than '../../src/...' which is a bit ugly.

I might have missed some that should be there, this is what came to mind.

### DIFF
--- a/include/nodes/Node
+++ b/include/nodes/Node
@@ -1,0 +1,2 @@
+#include "../../src/Node.hpp"
+
--- a/include/nodes/NodeGraphicsObject
+++ b/include/nodes/NodeGraphicsObject
@@ -1,0 +1,2 @@
+#include "../../src/NodeGraphicsObject.hpp"
+


paceholder:
I did not see the situations where I would need the direct access to that classes. All user is supposed to do is to define models

paceholder:
I think the whole approach I implemented here is wrong. It is convenient to build the library with examples. But in a distributed form (libs and includes on 3rd party machine) that headers will not work. Probably, I should implement the install step in CMake and move all the headers to the `include` directory

russelltg:
What I do in most of my libraries is to have the source tree resemble what
is  actually installed by the library. This makes it really easy to create
install steps and allows for easy in-source builds.

Something like:

```
include/
    nodes/
        public *.hpp files
src/
    private *.hpps
    *.cpps

```

This is nice because then you can just use the `target_include_directories`
cmake function to make sure that all dependencies have the right include
directories and don't include what they shouldn't. Also the install step is
just copy libraries and the include directory.

paceholder:
Yeah, that looks good. I've just checked: Qt does the same. I will re-work the source tree soon.

paceholder:
Closing this one in favor of #18

PR #21, opened 2016-11-29
Add the uniqueModel in DataModelRegistry::registerModel to be an optional parameter

russelltg:
This way the custom metadata that is needed for runtime defined types can be added optionally. (I need this for my uses)


paceholder:
Yeah, sorry, I've removed your previous code trying to make interfaces simpler

PR #22, opened 2016-11-29
Expose another createConnection with both nodes

russelltg:
createConnection currently only supports one-sided creations, this has two. 
It makes custom deserialization much easier than before.


paceholder:
Merged manually

PR #24, opened 2016-12-01
Hash-based connection color for each NodeDataType

paceholder:


CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 3.2)
 
 project(NodeEditor)
 
-option(BUILD_EXAMPLES ON)
+option(BUILD_EXAMPLES "ON" ON)
```
paceholder: Looks like an bug in CMake. Without this text label the option is always OFF.

PR #28, opened 2016-12-12
Make Connection a QObject so it can have a updated signal

russelltg:
This makes it possible for it to update some underlying data model in real time.

---
This change is [Reviewable](https://reviewable.io/reviews/paceholder/nodeeditor/28)

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -205,6 +205,8 @@ setNodeToPort(std::shared_ptr<Node> node,
     _inPortIndex = portIndex;
 
   _connectionState.setNoRequiredPort();
+  
+  updated(*this);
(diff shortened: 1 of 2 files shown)


russelltg:
This might be a bit controversial because I'm not sure if `Connection` is supposed to be a `QObject` but I didn't find a cleaner way to do this--suggestions are welcome :);

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/paceholder/nodeeditor/28#-:-KYko3rLnprfqSnpFNtw:bdyw2a2)*

paceholder:
Well, as long as we are working in the Qt environment, I see no harm in adding another QObject.

PR #36, opened 2016-12-31
Add miscalenous changes

russelltg:
* Have the signals yield std::shared_ptr if callees want ownership
* Remove a call to updated
* Remove some resize calls that seem to be misplaced

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -220,7 +220,6 @@ removeFromNodes() const
   if (auto n = _outNode.lock())
     n->nodeState().eraseConnection(PortType::Out, _outPortIndex, id());
   
-  updated(*this);
 }
 
 
--- a/src/Node.cpp
+++ b/src/Node.cpp
@@ -177,9 +177,6 @@ propagateData(std::shared_ptr<NodeData> nodeData,
 {
   _nodeDataModel->setInData(nodeData, inPortIndex);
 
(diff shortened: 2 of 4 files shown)


src/Connection.cpp
```diff
@@ -220,7 +220,6 @@ removeFromNodes() const
   if (auto n = _outNode.lock())
     n->nodeState().eraseConnection(PortType::Out, _outPortIndex, id());
   
-  updated(*this);
```
russelltg: I don't think updated should go in here because this is always right before it is deleted, not exactly an update

src/Node.cpp
```diff
@@ -177,9 +177,6 @@ propagateData(std::shared_ptr<NodeData> nodeData,
 {
   _nodeDataModel->setInData(nodeData, inPortIndex);
 
-  _nodeGeometry.recalculateSize();
-  _nodeGraphicsObject->setGeometryChanged();
-  _nodeGraphicsObject->update();
```
russelltg: I think this is misplaced because think about this scenario:
- Node is deleted,
- The NodeDataModel is deleted
- An empty shared ptr is passed to propagateData
- recalculateSize needs the NodeDataModel, so segfault!

(This happened to me)

paceholder: honestly, I do not remember why this code is placed here. Maybe by mistake

PR #37, opened 2017-01-02
Have the connection signal be non-const

russelltg:
Also make `NodeGraphicsObject::NodeGraphcisObject` take in a `const std::shared_ptr<>&` to allow rvalues.


russelltg:
The first thing was resolved by #38

PR #38, opened 2017-01-02
Make Nodes uniquely owned by FlowScene instead of using shared_ptr

russelltg:
Nodes were stored in `shared_ptr` objects, which doesn't make sense
because `FlowScene` should be the only owner of Nodes.

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -4,6 +4,7 @@
 #include <math.h>
 
 #include <QtWidgets/QtWidgets>
+#include <qcompilerdetection.h>
 
 #include "Node.hpp"
 #include "FlowScene.hpp"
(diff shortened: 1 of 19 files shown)


src/Connection.cpp
```diff
@@ -4,6 +4,7 @@
 #include <math.h>
 
 #include <QtWidgets/QtWidgets>
+#include <qcompilerdetection.h>
```
russelltg: For Q_UNUSED()

src/Connection.cpp
```diff
@@ -254,8 +252,15 @@ connectionGeometry()
   return _connectionGeometry;
 }
 
+ConnectionGeometry const&
```
russelltg: For `ConnectionPainter::paint` which now takes in a `Connection const&` so there needs to be const access to the geometry

src/Connection.cpp
```diff
@@ -294,14 +300,15 @@ getNode(PortType portType)
       // not possible
       break;
   }
+  Q_UNREACHABLE();
```
russelltg: This will add basically an `assert(false)` if execution gets here

paceholder:
Hi! I agree, this makes sense. I was on vacation and couldn't answer.

PR #46, opened 2017-01-20
Runtime validation and error reporting

LandonJerre:
Extension to the library to provide a way to handle internal node errors in a more visible way. Nodes now have a chance to implement a kind of self validation mechanism, and report to the scene if the operation they representing is not possible on the data they contain. (Example in the calculator sample, if the division node gets a 0 value on the divisor port, that now results in a validation error.) In case of an error, the reported error message if displayed on the invalid node.
errormsg
Additionally there's a minor modification to create an easy way to iterate through all the existing nodes in the scene.

---
This change is [Reviewable](https://reviewable.io/reviews/paceholder/nodeeditor/46)

### DIFF
--- a/examples/calculator/DivisionModel.hpp
+++ b/examples/calculator/DivisionModel.hpp
@@ -13,6 +13,9 @@
 /// In this example it has no logic.
 class DivisionModel : public MathOperationDataModel
 {
+private:
+  bool isModelValid = true;
+  QString modelValidationError = QString("");
 public:
 
   virtual
@@ -45,7 +48,7 @@ public:
       default:
         break;
     }
-	return QString("");
+	  return QString("");
   }
   
   QString
@@ -56,6 +59,14 @@ public:
   clone() const override
   { return std::make_unique<DivisionModel>(); }
 
+  virtual
+  bool
+  isValid() const { return isModelValid; }
+
+  virtual
+  QString
+  errorMessage() const { return modelValidationError; }
+
 public:
 
   void
@@ -65,7 +76,13 @@ public:
   }
 
 private:
-
+  void
+  setValidationState(bool isValid, const QString &msg)
+  {
+    isModelValid = isValid;
+    modelValidationError = msg;
+  }
+  
   void
   compute() override
   {
@@ -74,14 +91,24 @@ private:
     auto n1 = _number1.lock();
     auto n2 = _number2.lock();
 
-    if (n1 && n2 && (n2->number() != 0.0))
+    if (n2 && (n2->number() == 0.0))
     {
-      _result = std::make_shared<NumberData>(n1->number() /
-                                             n2->number());
+      setValidationState(false, 
+        QString("Division by zero error"));
     }
     else
     {
-      _result.reset();
+      setValidationState(true,
+        QString(""));
+      if (n1 && n2 && (n2->number() != 0.0))
+      {
+        _result = std::make_shared<NumberData>(n1->number() /
+          n2->number());
+      }
+      else
+      {
+        _result.reset();
+      }
     }
 
     emit dataUpdated(outPortIndex);
--- a/examples/calculator/DivisionModel.hpp
+++ b/examples/calculator/DivisionModel.hpp
@@ -48,7 +48,7 @@ public:
       default:
         break;
     }
-	  return QString("");
+    return QString("");
   }
   
   QString
@@ -59,13 +59,13 @@ public:
   clone() const override
   { return std::make_unique<DivisionModel>(); }
 
-  virtual
   bool
-  isValid() const { return isModelValid; }
+  isValid() const override 
+  { return isModelValid; }
 
-  virtual
   QString
-  errorMessage() const { return modelValidationError; }
+  errorMessage() const override
+  { return modelValidationError; }
 
 public:
 
@@ -76,13 +76,7 @@ public:
   }
 
 private:
-  void
-  setValidationState(bool isValid, const QString &msg)
-  {
-    isModelValid = isValid;
-    modelValidationError = msg;
-  }
-  
+
   void
   compute() override
   {
@@ -93,14 +87,14 @@ private:
 
     if (n2 && (n2->number() == 0.0))
     {
-      setValidationState(false, 
-        QString("Division by zero error"));
+      isModelValid = false;
+      modelValidationError = QString("Division by zero error");
     }
     else
     {
-      setValidationState(true,
-        QString(""));
-      if (n1 && n2 && (n2->number() != 0.0))
+      isModelValid = true;
+      modelValidationError = QString("");
+      if (n1 && n2)
       {
         _result = std::make_shared<NumberData>(n1->number() /
           n2->number());
--- a/examples/calculator/DivisionModel.hpp
+++ b/examples/calculator/DivisionModel.hpp
@@ -13,9 +13,6 @@
 /// In this example it has no logic.
 class DivisionModel : public MathOperationDataModel
 {
-private:
-  bool isModelValid = true;
-  QString modelValidationError = QString("");
 public:
 
   virtual
@@ -59,14 +56,6 @@ public:
   clone() const override
   { return std::make_unique<DivisionModel>(); }
 
-  bool
-  isValid() const override 
-  { return isModelValid; }
-
-  QString
-  errorMessage() const override
-  { return modelValidationError; }
-
 public:
 
   void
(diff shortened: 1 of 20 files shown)


examples/calculator/DivisionModel.hpp
```diff
@@ -45,7 +48,7 @@ class DivisionModel : public MathOperationDataModel
       default:
         break;
     }
-	return QString("");
+	  return QString("");
```
russelltg: 2 for indentation, not tab+2 spaces

examples/calculator/DivisionModel.hpp
```diff
@@ -56,6 +59,14 @@ class DivisionModel : public MathOperationDataModel
   clone() const override
   { return std::make_unique<DivisionModel>(); }
 
+  virtual
+  bool
+  isValid() const { return isModelValid; }
```
russelltg: `override`

examples/calculator/DivisionModel.hpp
```diff
@@ -56,6 +59,14 @@ class DivisionModel : public MathOperationDataModel
...
+
+  virtual
+  QString
+  errorMessage() const { return modelValidationError; }
```
russelltg: again `override`

examples/calculator/DivisionModel.hpp
```diff
@@ -65,7 +76,13 @@ class DivisionModel : public MathOperationDataModel
   }
 
 private:
-
+  void
+  setValidationState(bool isValid, const QString &msg)
```
russelltg: Is this function needed? It just manipulates private member variables and doesn't do anything special....

examples/calculator/DivisionModel.hpp
```diff
@@ -74,14 +91,24 @@ class DivisionModel : public MathOperationDataModel
...
-                                             n2->number());
+      setValidationState(false, 
+        QString("Division by zero error"));
     }
     else
     {
-      _result.reset();
+      setValidationState(true,
+        QString(""));
+      if (n1 && n2 && (n2->number() != 0.0))
```
russelltg: The last condition will never be true because it's checked on L94

LandonJerre:
Thanks for the feedback, I fixed the things you mentioned (I probably forgot to double check this file before I sent the pull request). On top of this there was a minor bug in the drawing logic that now got fixed: if you check the calculator example, the result node doesn't resize correctly if its value changes from empty to some number. (If you pull the mouse over it, that forces a refresh.) I already implemented a solution that handled this kind of size change, but it only ran if the validation state changed, so I removed the condition check from it to handle all kinds of data dependent size change. Theoretically recalculation could be a teeny bit slower because of this, but I think the size of the graph becomes unmanagable for the endusers much before this slowdown could be significant. (Although a check could be implemented, so that we only force an update when the size really changed.)

russelltg:
:+1: LGTM

russelltg:
I actually just had an idea. If `isValid` returns true and the error message is set, that could be a warning, displayed in yellow. That would be cool and useful (at least for me...)

russelltg:
Reviewed 7 of 9 files at r1, 2 of 2 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/FlowScene.cpp, line 224 at r2](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwjBSF3Able0zW_GH:-KbCwjBSF3Able0zW_GI:blsnjk3) ([raw file](https://github.com/paceholder/nodeeditor/blob/bfda8fef7fa94614e784dfac01e5886db19d9fa9/src/FlowScene.cpp#L224)):*
> ```cpp
>   for (const auto& _node : _nodes)
>   {
> 	visitor(_node.second.get());
> ```

2 tabs for indentation

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/paceholder/nodeeditor/46)*

LandonJerre:
Good idea, but instead of checking the message for content, I've changed isValid into returning an enum value. It's clearer, and easily extensible to handle additional validation states. (Multiple different warning levels for example.) I'm in the process of refactoring the entire branch to use this consistently, so a second full review after the commit would probably be a good idea.

LandonJerre:
The end result visually:
eg
The travis build magically failed, looking at the log it looks like it's not a compilation error.

LandonJerre:
Review status: 2 of 19 files reviewed at latest revision, 6 unresolved discussions, some commit checks broke.

---

*[examples/calculator/DivisionModel.hpp, line 51 at r1](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwNR0AwZ1egAEzb-i-r1-51:-KbFtdkXBbW_WYpk2H7G:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/340c20c00ca1442a2c68a427c10ef4fe81469cf1/examples/calculator/DivisionModel.hpp#L51)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

2 for indentation, not tab+2 spaces
</blockquote></details>

Done.

---

*[examples/calculator/DivisionModel.hpp, line 64 at r1](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwNR0AwZ1egAEzb-i-r1-64:-KbFtd6cHTXgLLXi1otR:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/340c20c00ca1442a2c68a427c10ef4fe81469cf1/examples/calculator/DivisionModel.hpp#L64)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

`override`
</blockquote></details>

Done.

---

*[examples/calculator/DivisionModel.hpp, line 68 at r1](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwNR0AwZ1egAEzb-i-r1-68:-KbFtcgCV6tkcM1RRWip:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/340c20c00ca1442a2c68a427c10ef4fe81469cf1/examples/calculator/DivisionModel.hpp#L68)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

again `override`
</blockquote></details>

Done.

---

*[examples/calculator/DivisionModel.hpp, line 80 at r1](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwNR0AwZ1egAEzb-i-r1-80:-KbFtbmpjyK1z0UOJ6RP:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/340c20c00ca1442a2c68a427c10ef4fe81469cf1/examples/calculator/DivisionModel.hpp#L80)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

Is this function needed? It just manipulates private member variables and doesn't do anything special....
</blockquote></details>

Done.

---

*[examples/calculator/DivisionModel.hpp, line 103 at r1](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwNR0AwZ1egAEzb-i-r1-103:-KbFt_kONVJB6wfmxmQI:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/340c20c00ca1442a2c68a427c10ef4fe81469cf1/examples/calculator/DivisionModel.hpp#L103)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

The last condition will never be true because it's checked on L94
</blockquote></details>

Done.

---

*[src/FlowScene.cpp, line 224 at r2](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-KbCwjBSF3Able0zW_GH:-KbFtfwCHzV8TbHLay_G:b-896fix) ([raw file](https://github.com/paceholder/nodeeditor/blob/bfda8fef7fa94614e784dfac01e5886db19d9fa9/src/FlowScene.cpp#L224)):*
<details><summary><i>Previously, russelltg (Russell Greene) wrote…</i></summary><blockquote>

2 tabs for indentation
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/paceholder/nodeeditor/46)*

LandonJerre:
Travis build is working again, it looks like the qt5.7 repository path changed, correcting it seemingly fixed the issue.

russelltg:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/paceholder/nodeeditor/46#-:-KbI4Z59edKHqSy66Es3:bnfp4nl)*

russelltg:
I really like this, I hope @paceholder merges this 👍

paceholder:
I really like the idea and was thinking of implementing something like that. 

The only concern I have is the visualization: it looks heavy and changes geometry.

Maybe it is worth to show the incorrect input just by highlighting the input circle with red?
And show "no data propagation" with graying out the nodes without the valid data?

So, I will merge it for now and will think how to improve the graphics.

LandonJerre:
> The only concern I have is the visualization: it looks heavy and changes geometry.

Idea: if the node is in a validation state that is not Valid, we put a small square in one of the nodes corner with a ! in it, and a yellow or red background color. If the user clicks on this square a bigger square pops out next to the node, containing the validation message. This would allow longer messages without serious change to the nodes visuals it self. Would this solve the problem? I've planned to implement something like this for help info/documentation on nodes, if I manage to get around doing it, it could easily be repurposed for the validation messages too. 

> Maybe it is worth to show the incorrect input just by highlighting the input circle with red?
And show "no data propagation" with graying out the nodes without the valid data?

This could be done, and it would look good, but for this to work, the node validation part must be rewritten. The current solution validates the node as a whole, not the individual ports. Changing the validationState and validationMessage functions to have a port index and port type param should do the trick from an interface perspective.

PR #50, opened 2017-01-26
Context menu filter

LandonJerre:
Minor addition to the node insert context menu, to make it's contents real time filterable. With a bunch of different node types, this could make the editor much easier to use.
filter


russelltg:
Haha, I was actually just implementing this exact feature yesterday (you did a better job though).

👍 

It can get tedious when you have ~40 node types in chigraph (so far)

russelltg:
LGTM

PR #51, opened 2017-01-28
Fix segfault in NodeScene

russelltg:
The CGO is requested in `NodeGraphicsObject::moveConnections` with the call stack:
```
Node::onDataUpdated
Connection::propagateData
Node::onDataUpdated
```

Which is called in createConnection


LandonJerre:
I fixed this exact thing yesterday in my fork, in exactly the same way you did. :)

paceholder:
Thanks guys!

PR #59, opened 2017-03-03
it will compile on c++11 and Qt 5.2

facontidavide:
Hi, 

the main source of incompatibility were std::make_unique and QMarginsF.

I also add to fix something in the macros (I am not sure why that was necessary).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,14 +15,14 @@ matrix:
           packages:
             - g++-4.8
     - compiler: clang
-      env: CXX=clang++-3.5 CC=clang-3.5
+      env: CXX=clang++-3.6 CC=clang-3.6
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.5
+            - llvm-toolchain-precise-3.6
           packages:
-            - clang-3.5
+            - clang-3.6
 
 
 #branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,22 +7,22 @@ language: none
 matrix:
   include:
     - compiler: gcc
-      env: CXX=g++-5 CC=gcc-5 QT_BASE=57
+      env: CXX=g++-4.8 CC=gcc-4.8
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
           packages:
-            - g++-5
+            - g++-4.8
     - compiler: clang
-      env: CXX=clang++-3.7 CC=clang-3.7 QT_BASE=57
+      env: CXX=clang++-3.5 CC=clang-3.5
       addons:
         apt:
           sources:
             - ubuntu-toolchain-r-test
-            - llvm-toolchain-precise-3.7
+            - llvm-toolchain-precise-3.5
           packages:
-            - clang-3.7
+            - clang-3.5
 
 
 #branches:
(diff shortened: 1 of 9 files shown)


paceholder:
I will not comment each file, just write all remarks here:

1. `make_unique` must be in `std::` namespace. That would a) help to avoid lots of corrections b) compile in c++14 without any problems.

2. The alternative implementation for `make_unique` must be in a separate file `filename`

3. The alternative implementation must be guarded by `#ifdef` and enabled just
    if the `standard number < 14`. There are ways to check the currently enable standard version.

src/ConnectionStyle.cpp
```diff
@@ -64,7 +64,7 @@ setConnectionStyle(QString jsonText)
         auto colorArray = valueRef.toArray(); \
         std::vector<int> rgb; rgb.reserve(3); \
         for (auto it = colorArray.begin(); it != colorArray.end(); ++it) { \
-          rgb.push_back(it->toInt()); \
+          rgb.push_back((*it).toInt()); \
```
paceholder: Is it related to the C++14 or Qt version?

src/Node.cpp
```diff
@@ -55,11 +55,10 @@ save() const
 
   nodeJson["model"] = _nodeDataModel->save();
 
-  nodeJson["position"] = QJsonObject
-  {
-    { "x" , _nodeGraphicsObject->pos().x() },
-    { "y" , _nodeGraphicsObject->pos().y() }
-  };
```
paceholder: It wasn't compilable with C++11 right?

facontidavide: yes, you are right

facontidavide:
Ok, I cleaned it up, thanks for the suggestions

src/ConnectionStyle.cpp
facontidavide: I sincerely have no idea. It just works if I do that.

src/FlowViewStyle.cpp
```diff
@@ -56,7 +56,7 @@ setStyle(QString jsonText)
       auto colorArray = valueRef.toArray(); \
       std::vector<int> rgb; rgb.reserve(3); \
       for (auto it = colorArray.begin(); it != colorArray.end(); ++it) { \
-        rgb.push_back(it->toInt()); \
+        rgb.push_back((*it).toInt()); \
```
facontidavide: One of those mysteries of the universe that we will never know

paceholder:
The windows build is failing because MSVC does not rely on the provided standard version. It just gives all the functionality for the current release. std::make_unique is implemented in the MSVC used by appveyor CI.

I used to fix this issue earlier with

    #include "Compiler.hpp"		

    #if !defined(_MSC_VER) ||                     \		
       (defined(_MSC_VER) && (_MSC_VER < 1900))

    // here your implementation for std::make_unique

    #endif

facontidavide:
Everything looks fine now :)

.travis.yml
```diff
@@ -31,14 +31,10 @@ matrix:
...
 before_install:
-  #- if [ "$QT_BASE" = "54" ]; then sudo add-apt-repository ppa:beineri/opt-qt542-trusty -y; fi
-  - if [ "$QT_BASE" = "57" ]; then sudo add-apt-repository ppa:beineri/opt-qt571-trusty -y; fi
   - sudo apt-get update -qq
 
 install:
-  - sudo apt-get install build-essential libgl1-mesa-dev
-  #- if [ "$QT_BASE" = "54" ]; then sudo apt-get install -qq qt54base; source /opt/qt54/bin/qt54-env.sh; fi
-  - if [ "$QT_BASE" = "57" ]; then sudo apt-get install -qq qt57base; source /opt/qt57/bin/qt57-env.sh; fi
+  - sudo apt-get install build-essential libgl1-mesa-dev qtbase5-dev
```
paceholder: Interesting, it did now work earlier without explicit installation of Qt

.travis.yml
```diff
@@ -7,22 +7,22 @@ language: none
 matrix:
   include:
     - compiler: gcc
```
paceholder: Why the versions are downgraded below?

facontidavide: Because the goal of this PR is to make it compile with default libraries in Trusty, when __build-essential__ and __qtbase5-dev__ are installed.

.travis.yml
facontidavide: well, qt __IS__ explicitily installed, just the default one that is 5.2.1

paceholder: Oh, I see. Sorry, I did it quite a while ago. Looks that I needed the latest Qt.

Issue #54, opened 2017-02-09
Crash on Load of a new node graph


mtbrobotanist:
Hi, awesome product with one small issue:

The Node editor calculator example crashes when loading a second .flow file after an initial load. loading a new graph in its place causes an memory access violation whenever FlowScene::load() calls _connections.clear(); 

I'm using Visual Studio 2015 Community on Windows 10 with Qt 5.7.

paceholder:
Hi! Thanks for reporting an issue, I will review it in the nearest time.

PR #60, opened 2017-03-07
Various fixes and additions

LandonJerre:
- Fixed a crash, that happened when the user tried to delete multiple selected nodes and connections
- Fixed a crash, that happened when the user tried to load a scene from file into a non-empty scene
- Added functions to FlowScene for node position manipulation, and size readback. Before this, this data were only accessible through types that are not exported from the library, but on the other hand moving nodes in code is useful for automatic graph generation. (Like generating a basic graph that the user edits, instead of giving them an empty scene.)
- Added functions to Flowscene for iterating through the existing nodes. iterateOverNodeData just iterates through the nodes, passing the node data to the visitor lambda. iterateOverNodeDataDependentOrder iterates through the nodes in a specific order, it always visits a nodes all input nodes, before visiting the node itself. The latter is useful for codegeneration from the graph, where the order of the generated statements are important, and statements that feed other statements must be before them.
- Redid the context menu to be usable with a larger number of node types. Instead of dumping all the types into the menu, now all of them are put into a treewidget, under the categories that are passed during type registration. (A sideeffect is, that the filtering is noticably faster now, with the old way, filtering in the list on the picture froze the ui for a few moment after the first character inputed.) To illustrate the problem, and the solution:
comp

### DIFF
--- a/src/DataModelRegistry.hpp
+++ b/src/DataModelRegistry.hpp
@@ -50,7 +50,7 @@ public:
 
   template<typename ModelType, bool TypeConverter = false>
   void
-  registerModel(std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>(), const QString& category = "Nodes")
+  registerModel(std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>(), QString const &category = "Nodes")
   {
     static_assert(std::is_base_of<NodeDataModel, ModelType>::value,
                   "Must pass a subclass of NodeDataModel to registerModel");
@@ -89,7 +89,7 @@ public:
   //Parameter order alias, so a category can be set without forcing to manually pass a model instance
   template<typename ModelType, bool TypeConverter = false>
   void
-  registerModel(const QString& category, std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>())
+  registerModel(QString const &category, std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>())
   {
     registerModel<ModelType, TypeConverter>(std::move(uniqueModel), category);
   }
@@ -107,8 +107,8 @@ public:
   categories() const;
 
   std::unique_ptr<NodeDataModel>
-  getTypeConverter(const QString & sourceTypeID, 
-                   const QString & destTypeID) const;
+  getTypeConverter(QString const &sourceTypeID,
+                   QString const &destTypeID) const;
 
 private:
 
--- a/src/DataModelRegistry.hpp
+++ b/src/DataModelRegistry.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <unordered_map>
+#include <set>
 #include <memory>
 
 #include <QtCore/QString>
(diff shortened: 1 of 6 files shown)


LandonJerre:
The second point probably takes care of issue #54

src/DataModelRegistry.hpp
```diff
@@ -47,7 +50,7 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 
   template<typename ModelType, bool TypeConverter = false>
   void
-  registerModel(std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>())
+  registerModel(std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>(), const QString& category = "Nodes")
```
paceholder: minor: please follow the style  `Type const &`

src/FlowScene.cpp
```diff
@@ -230,6 +230,98 @@ iterateOverNodes(std::function<void(Node*)> visitor)
...
+  for (const auto& _node : _nodes)
+  {
+    visitor(_node.second->nodeDataModel());
+  }
+}
+
+
+void
+FlowScene::
+iterateOverNodeDataDependentOrder(std::function<void(NodeDataModel*)> visitor)
```
paceholder: I would simplify and restructure the first part as follows:

```
auto isNodeWithZeroInputs = [](Node const & node)
{
  bool result = true;

  auto model = node->nodeDataModel();

  for(std::size_t i = 0; i < model->nPorts(PortType::In); ++i)
  {
    auto connections = _node->nodeState().connections(PortType::In, i);

    if (!connections.empty())
    {
      result = false;
      break;
    }
  }

  return result;
};


std::size_t nodeIndex = 0;

for (auto const & idNodePair: _nodes)
{
  auto const &node = idNodePair.second;

  if (isNodeWithZeroInputs(node))
  {
    nodeIndexMap[node->id()] = nodeIndex;

    auto model = node->nodeDataModel();

    visitor(model);
  }

  ++nodeIndex;
}
```

src/FlowScene.cpp
```diff
@@ -230,6 +230,98 @@ iterateOverNodes(std::function<void(Node*)> visitor)
...
+      auto model = _node.second->nodeDataModel();
+      for (size_t i = 0; i < model->nPorts(PortType::In); ++i)
+      {
+        auto connections = _node.second->nodeState().connections(PortType::In, i);
+
+        for (auto& conn : connections)
+        {
+          if (nodeIndexMap.find(conn.second->getNode(PortType::Out)->id()) == nodeIndexMap.end())
+          {
+            goto skip_node_dep;
```
paceholder: I think it must be changed as above. It is possible to do the job without `goto`.

paceholder:
Thanks for this great add-on. I made a couple of comments

LandonJerre:
Fixed the code style issue (I made that mistake in another function prior, so fixed that as well), refactored iterateOverNodeDataDependentOrder to get rid of goto, plus I did a few minor simplifications on it, so hopefully it's much easier to read it now.

paceholder:
I will merge it. Please, next time, if possible do several independent pull-requests or commits for each implemented feature.

LandonJerre:
Sorry about that, I realized after I finished this, that I should've separated my changes. All the stuff I did/am doing right now are separated logically, so the next pull requests won't be this messy hopefully.

PR #61, opened 2017-03-09
Compilation fix for VS2013

LandonJerre:
Changed the make_unique 'emulation' conditions, so that it doesn't match VS2013. (VS2013 already has make_unique in std, the standard compliance of VS versions are all over the place.)

### DIFF
--- a/src/make_unique.hpp
+++ b/src/make_unique.hpp
@@ -4,15 +4,17 @@
 
 
 #if (!defined(_MSC_VER) && (__cplusplus < 201300)) || \
-    ( defined(_MSC_VER) && (_MSC_VER < 1900))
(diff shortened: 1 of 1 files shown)


russelltg:
Might not be a bad idea to add msvc 2013 to CI to catch things like this in the future

LandonJerre:
Well even with this fixed, the code won't compile out of the box with VS2013, because it doesn't support defaulting move constructors and move assignment operators. DataModelRegistry.hpp line 42 and line 46-47 are the only places where this matters, so for now I just comment these lines out locally, to hack around the problem. The proper fix probably would be to implement these two functions by hand, instead of defaulting them.

paceholder:
There is a hack

http://stackoverflow.com/questions/13883824/make-unique-does-not-compile

PR #62, opened 2017-03-09
Save/load graph from memory

LandonJerre:
Added saveToMemory/loadFromMemory functions, so the users can implement their own method of getting the data, that will be loaded into the graph. The original save/load are refactored to use the new functions.


paceholder:
Oh, I thought I've merged it already.

PR #64, opened 2017-03-10
Add corner decorators

russelltg:
This allows people to add `QIcon`s to the corners of nodes. 

Here's a screenshot:

image

Users just need to override the `cornerDecorator` function in `NodeDataModel`

### DIFF
--- a/src/NodeDataModel.hpp
+++ b/src/NodeDataModel.hpp
@@ -3,6 +3,7 @@
 #include <memory>
 
 #include <QtWidgets/QWidget>
+#include <QIcon>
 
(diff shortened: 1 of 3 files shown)


paceholder:
We cannot answer each specific need by adding more and more functions to the model.
This problem could be solved more generally:

- A model would have the function `NodePaintDelegate* Model::pointerDelegate() const` (could be `nullptr`)
- User is to implement `NodePainterDelegate::paint(QPainter*, QRect)`

russelltg:
That's a better idea. I can implement this.

paceholder:
Thank you!

PR #65, opened 2017-03-10
Add selectedNodes() function to FlowScene

russelltg:
I also had to add `type()` override to `NodeGraphicsObject` to allow for [qgraphicsitem_cast](http://doc.qt.io/qt-5/qgraphicsitem.html#qgraphicsitem_cast) to work

### DIFF
--- a/src/FlowScene.cpp
+++ b/src/FlowScene.cpp
@@ -351,6 +351,24 @@ connections() const
   return _connections;
 }
 
+std::vector<Node*>
+FlowScene::selectedNodes() const {
+  QList<QGraphicsItem*> graphicsItems = selectedItems();
+
+  std::vector<Node*> ret;
+  ret.reserve(graphicsItems.size());
+
+  for (QGraphicsItem* item : graphicsItems) {
+    auto ngo = qgraphicsitem_cast<NodeGraphicsObject*>(item);
+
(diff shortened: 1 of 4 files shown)


src/FlowScene.cpp
```diff
@@ -351,6 +351,24 @@ connections() const
...
 
+std::vector<Node*>
+FlowScene::selectedNodes() const {
+  QList<QGraphicsItem*> graphicsItems = selectedItems();
+
+  std::vector<Node*> ret;
+  ret.reserve(graphicsItems.size());
+
+  for (QGraphicsItem* item : graphicsItems) {
+    auto ngo = qgraphicsitem_cast<NodeGraphicsObject*>(item);
```
paceholder: is it faster than `dynamic_cast` ?

russelltg: I think so. `dynamic_cast` uses RTTI, where `qgraphicsitem_cast` just checks type with a virtual function.

russelltg: Did some research, found this in some qt textbook:
image

paceholder: Well, it is Qt's decision. I just wonder what would they do with dozens of different graphical items. It must be a mess in defining the `type()` function and enums correctly

PR #67, opened 2017-03-14
Add onPaint callback to NodeDataModel for custom painting

russelltg:
I also exposed NodeGeometry to the public API, which had to be done because it'd be hard to expose painting things without knowing where you are. 

If you have a better way to do this, I'm all ears.

I decided to go with a simple virtual function because it does basically the same thing as a delegate, but it's simpler and more concise.

### DIFF
--- a/include/nodes/NodeGeometry
+++ b/include/nodes/NodeGeometry
@@ -1,0 +1,2 @@
+#include "../../src/NodeGeometry.hpp"
+
--- a/include/nodes/NodePainterDelegate
+++ b/include/nodes/NodePainterDelegate
@@ -1,0 +1,1 @@
+#include "../../src/NodePainterDelegate.hpp"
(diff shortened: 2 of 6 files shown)


paceholder:
This would definitely do the job but, sorry, it is still mixing data model and painting mechanism.
As I mentioned earlier, the correct way to do it is to return a `NodePainterDelegate` from the model.
`Model` defines the behavior: to do some custom painting or not to do.
`NodePainterDelegate` defines how this behavior is implemented.

`Model` does not paint anything the same way it does not paint the basic node rectangle.

Qt does pretty much the same for its `QTreeView` and `QAbstractItemModel` and related classes. To paint an item differently we'd use a delegate class there.

russelltg:
I added a new `NodePainterDelegate` class
Do you think it'd be better if `NodePainterDelegate` was defined as a `std::function<...>`? It would be less verbose, and possibly more flexible for clients.

russelltg:
Thanks, it is a bit cleaner that way.

PR #70, opened 2017-03-21
Use actions in FlowView instead of built-in shortcuts

russelltg:
This changes no behavior, just adds flexibility.

This allows for

1. Changing shortcuts/removing shortcuts
2. Adding these actions to a KActionCollection
3. Easily attaching them to buttons


paceholder:
That's good, thanks

PR #82, opened 2017-04-12
fixing a problem with the Z level that affects proxywidgets too

facontidavide:
nodeeditor_zvalue

Hi,

 am having a problem with a CombBox hidden behind other NodeGraphiObjects.
I realized that in general the order in Z might be annoying in some cases because is related to the creation order.

This PR proposes a solution to this problem.

NOTE: if you prefer, the increase in Zvalue can be done in **hoverEnterEvent** instead of **mousePressEvent**. I would personally prefer hoverEnterEvent.

### DIFF
--- a/src/NodeGraphicsObject.cpp
+++ b/src/NodeGraphicsObject.cpp
@@ -314,14 +314,18 @@ void
 NodeGraphicsObject::
 hoverEnterEvent(QGraphicsSceneHoverEvent * event)
 {
-  // bring all the nodes to background
-  for(auto& it: _scene.nodes() )
+  // bring all the colliding nodes to background
+  QList<QGraphicsItem *> overlapItems = this->collidingItems();
+
+  foreach (QGraphicsItem *item, overlapItems)
   {
-      Node* node = it.second.get();
-      node->nodeGraphicsObject().setZValue(0.0);
+      if (item->zValue() > 0.0 )
+      {
+          item->setZValue(0.0);
+      }
   }
-  // bring this node to front
-  setZValue(1.0);
+  // bring this node forward
+  this->setZValue(1.0);
 
   _node.nodeGeometry().setHovered(true);
   update();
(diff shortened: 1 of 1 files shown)


paceholder:
Hi Davide,

Thanks for this fix. I never tried to use some combo-boxes.

I would solve this problem little bit differently.

1. Make an enum in Node.hpp file
```
enum NodeZLevel
{
    DEFAULT_Z_LEVEL = 1,   // we need to check appropriate values here
    HOVERED_Z_LEVEL = 100
};
```

2. In `hoverEnterEvent` set node's z-level dynamically to `HOVERED_Z_LEVEL`
3. In `hoverLeaveEvent` set it back to `DEFAULT_Z_LEVEL`

Benefits:
- No global static variables
- No constantly growing z-levels


What do you think?

Best,
Dmitry

facontidavide:
this is the very first thing I tried, but it didn't work.
When you click the combobox, hoverLeaveEvent is called.
I also tried to tinker with the setZvalue of _proxywidget, but without success...

paceholder:
Ok, I did not know that. Maybe we should look into Qt's example?

http://doc.qt.io/qt-5/qtwidgets-graphicsview-diagramscene-example.html

Please look at the function implementation `bringToFront()`

facontidavide:
Alright, I changed the PR and, code wise, it is aesthetically more pleasant ;)
Take a look

facontidavide:
shall we move this one forward? ;)

paceholder:
Yes, sorry, could not encompass everything in a couple of evening hours after work :-) 

There is on issue with the current implementation: by touching all the Nodes in the cycle we would cause a massive repainting, basically we would repaint the whole scene. It is not a problem for 10 nodes, but who knows, what if we have 1000? I suppose we could avoid wasting of computational resources and use the trick proposed in the Qt docs (http://doc.qt.io/qt-5/qtwidgets-graphicsview-diagramscene-example.html)

Thanks

Dmitry

facontidavide:
I pushed the changes you suggested (more or less...) I honestly don't think it need to be more complicated than that.

paceholder:
Thanks, now I am happy ;)

PR #84, opened 2017-04-20
Allow specific node styles for each node

facontidavide:
Hi,
I have an application in which I want to customize individually the style of each node.
This PR allows that without changing the default behavior.

### DIFF
--- a/src/Node.hpp
+++ b/src/Node.hpp
@@ -13,6 +13,7 @@
 #include "NodeState.hpp"
 #include "NodeGeometry.hpp"
 #include "NodeData.hpp"
+#include "NodeGraphicsObject.hpp"
 #include "Serializable.hpp"
 
 namespace QtNodes
--- a/src/NodeDataModel.hpp
+++ b/src/NodeDataModel.hpp
@@ -9,7 +9,8 @@
 #include "Serializable.hpp"
 #include "NodeGeometry.hpp"
 #include "NodePainterDelegate.hpp"
-
+#include "NodeStyle.hpp"
+#include "StyleCollection.hpp"
 #include "Export.hpp"
 
 namespace QtNodes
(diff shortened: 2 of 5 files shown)


paceholder:
Wouldnt' it be better to return a custom style from the model?
For example, as QString or JSON. 
An empty custom style would mean that the default one should be applied

russelltg:
No, I think this way is better because then you loose the flexibility of
changing the style of just one node, and can only change the style of all
nodes of one type. For example, like if a user "bookmarked" a node, you'd
probably want to change the style of that one node, but not all nodes of
the same type.

paceholder:
You would still be able to change the style of one node even if the style data is stored in the model.

The style initialization would work in three stages:
1. Set global NodeStyle as in the `styles` example (or use default global style by doing nothing).
2. [optionally] In the Model constructor set "specific" model-defined Node style
3. [optionally] Modify model style depending on the model/node state 

The style usage would also work in two stages:
1. Check if the model possesses a specific style
2. If not, use global style from `StyleCollection::nodeStyle`

With the initially proposed approach it wouldn't be impossible to define a model-default style

facontidavide:
> With the initially proposed approach it wouldn't be possible to define a model-default style

Good point.
it looks weird to include some graphic style into de DataModel, but pragmatically speaking it would do what I want.

russelltg:
> You would still be able to change the style of one node even if the style data is stored in the model.

Right of course. Wasn't thinking about it right :).

facontidavide:
Alright, I applied the changes and rebased the branch. Take a look

paceholder:
Merged manually with tiny code style fixes.

paceholder:
> it looks weird to include some graphic style into de DataModel,

On one hand yes, it looks weird.

But on other hand, Qt does almost the same with 
`QVariant QAbstractItemModel::data(const QModelIndex &index, int role = Qt::DisplayRole) const`

There we can return QIcon for any abstract-item-based widget using the role == Qt::DecorationRole.

PR #90, opened 2017-04-25
allow the ability to lock a node/connection in the editor

facontidavide:
This PR allow the user to "lock" a node or connection, i.e. to prevent the user from moving and/or deleting it.


paceholder:
It would be nice to add a very simple example.

facontidavide:
yes, I will prepare this example  in the next days. Cheers

PR #105, opened 2017-08-09
Correct -Wall and -Wextra warnings for GCC7.1.1 and Clang++ 4.0.1

rkollataj:
We are using your project as a sub-module in CANdevStudio. As the project is build with -Wall and -Wextra flags compilation of your code produces few minor warnings - unused variables and order of initialisation. If you would like to enable these flags in your project also this PR will do it for you :-)

Signed-off-by: Remigiusz Kołłątaj <remigiusz.kollataj@mobica.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ set(CMAKE_POSITION_INDEPENDENT_CODE ON)
 
 set(CMAKE_CXX_STANDARD 14)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+add_definitions(-Wall -Wextra)
 
 #############################################################
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,15 +31,16 @@ find_package(Qt5 COMPONENTS
 add_definitions(${Qt5Widgets_DEFINITIONS})
 set(CMAKE_CXX_FLAGS "${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}")
 
-IF (MSVC)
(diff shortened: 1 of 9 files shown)


paceholder:
Hi Remigiusz,

your change is perfectly ok but the project should be compilable with all the compilers used in CI.

Maybe we can use this cmake code:
```
if(MSVC)
  # Force to always compile with W4
  if(CMAKE_CXX_FLAGS MATCHES "/W[0-4]")
    string(REGEX REPLACE "/W[0-4]" "/W4" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
  else()
    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /W4")
  endif()
elseif(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
  # Update if necessary
  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -pedantic -Wextra")
endif()
```

inspired by
https://stackoverflow.com/questions/2368811/how-to-set-warning-level-in-cmake

Regards,
Dmitry

rkollataj:
Oh right I forgot that MSVC doesn't like -Wextra. I am using something like below to make it work:

if(NOT MSVC)
    add_definitions(-Wall -Wextra)
else()
    add_definitions(/Wall)
endif()

Whichever works better for you. Do you want me to update PR with some of these?

paceholder:
Yes please, go ahead with your suggestion.

Regards,
Dmitry

PR #106, opened 2017-08-28
change _scene visibility to protected

ypnos:
Handy when deriving from FlowView to change behavior, e.g. implement own context menu or interaction with the scene from external widgets.


paceholder:
Pushed the commit with a protected member function:

```
protected:
    FlowScene * scene();
```

PR #107, opened 2017-08-28
Support inclusion of FlowView via Qt Designer

ypnos:
To be able to fit this widget in the typical .ui file workflow, we need to be able to construct it with a parent widget and set the scene after UI setup.

### DIFF
--- a/src/FlowView.cpp
+++ b/src/FlowView.cpp
@@ -29,9 +29,9 @@ using QtNodes::FlowScene;
 FlowView::
 FlowView(QWidget *parent)
   : QGraphicsView(parent)
-  , _scene(Q_NULLPTR)
   , _clearSelectionAction(Q_NULLPTR)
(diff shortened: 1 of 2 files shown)


ypnos:
Please note that although this is two commits they are not independent of each other, I just wanted to highlight the different changes. The introduction of `init()` is to stay compatible with C++03. Personally I would prefer to use a delegating constructor.

paceholder:
Thanks for the pull-request.

I think there is no need to stick to C++03 standard. The code base uses C++14 in the CMakeLists.txt and is being build with C++11 compilers in continuous integration.

Could you update it with the delegating constructors? Thanks

Dmitry

paceholder:
Could you replace `init()` with a delegating constructor?

ypnos:
Sorry for the delay, I am on it!

paceholder:
No problem, I was just pinging :)

PR #112, opened 2017-09-13
Modifications to allow custom context menus on node objects

tuvok:


src/FlowScene.hpp
```diff
@@ -112,6 +112,8 @@ class NODE_EDITOR_PUBLIC FlowScene
 
   void nodeHoverLeft(Node& n);
 
+  void nodeContextMenu(Node& n, const QPointF& pos);
```
paceholder: Sorry, where is the implementation of this function?

tuvok: It's in the signal: section, so the implementation is moc-generated

paceholder: Oh, indeed, I overlooked it.

PR #117, opened 2017-09-20
Update to use modelview philosophy

russelltg:
For a while, @paceholder and I have been talking about having a separation between the node rendering and the logic of a data flow framework. This brings that.

Use case: When you already have a data structure representing a flow scene, you can easily map an interface to it without keeping two data structures in sync.

Abstract:
This is inspired by Qt's model view infrastructure where there is a index for a particular item, a model (an abstract class) and a view (that takes in a pointer to a model).

New/Totally refactored classes:
`NodeIndex`: This is similar to a [`QModelIndex`](http://doc.qt.io/qt-5/qmodelindex.html), but more specific to nodes. It holds nothing but an ID (QUuid), an internal pointer, and a pointer to the model.

`FlowSceneModel`: This is the abstract model class. It defines a bunch of functions for reading and manipulating flow scenes. 

`DataFlowScene`: This roughly replaces `FlowScene`, and is actually a drop-in replacement. It subclasses from `FlowScene` and adds an implementation of `FlowSceneModel` that does exactly what this library used to do (data flow, with propagating data etc)

`ConnectionGraphicsObject`: This class has kinda been merged with `Connection`, and it now does the functionality for both. Basically, it handles all the backend scene stuff for rendering connections.

`Connection`: This is now entirely broken away from the rendering part of the library. It is now just part of the `DataFlowScene`.

`NodeGraphicsObject`/`Node`: Same deal as connections

`ConnectionID`: Just a way to identify connections in a map. 

Among plenty of other changes, but most users should just be able to change `FlowScene` to `DataFlowScene` and it'll work exactly as before. 

it's kinda a complex pull request. Sorry :/

### DIFF
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -26,6 +26,7 @@ using QtNodes::FlowScene;
 using QtNodes::NodeDataType;
 using QtNodes::NodeIndex;
 using QtNodes::PortIndex;
+using QtNodes::PortType;
 
 ConnectionGraphicsObject::
 ConnectionGraphicsObject(NodeIndex const& leftNode,
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -24,6 +24,7 @@ using QtNodes::Connection;
 using QtNodes::ConnectionID;
 using QtNodes::FlowScene;
 using QtNodes::NodeDataType;
+using QtNodes::NodeIndex;
 
 ConnectionGraphicsObject::
 ConnectionGraphicsObject(NodeIndex const& leftNode,
@@ -31,14 +32,14 @@ ConnectionGraphicsObject(NodeIndex const& leftNode,
                          NodeIndex const& rightNode,
                          PortIndex rightPortIndex,
                          FlowScene&       scene)
-  : _state(leftNode.isValid() ?
+  : _scene{scene}
+  , _state(leftNode.isValid() ?
            (rightNode.isValid() ?  PortType::None : PortType::In) :
            PortType::Out)
   , _leftNode{leftNode}
   , _rightNode{rightNode}
   , _leftPortIndex{leftPortIndex}
   , _rightPortIndex{rightPortIndex}
-  , _scene{scene}
 {
   _scene.addItem(this);
 
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -26,8 +26,14 @@ using QtNodes::FlowScene;
 using QtNodes::NodeDataType;
 
 ConnectionGraphicsObject::
-ConnectionGraphicsObject(const NodeIndex& leftNode, PortIndex leftPortIndex, const NodeIndex& rightNode, PortIndex rightPortIndex, FlowScene& scene)
-  : _state(leftNode.isValid() ? (rightNode.isValid() ? PortType::None : PortType::In) : PortType::Out)
+ConnectionGraphicsObject(NodeIndex const& leftNode,
+                         PortIndex leftPortIndex,
+                         NodeIndex const& rightNode,
+                         PortIndex rightPortIndex,
+                         FlowScene&       scene)
+  : _state(leftNode.isValid() ?
+           (rightNode.isValid() ?  PortType::None : PortType::In) :
+           PortType::Out)
   , _leftNode{leftNode}
   , _rightNode{rightNode}
   , _leftPortIndex{leftPortIndex}
@@ -47,14 +53,16 @@ ConnectionGraphicsObject(const NodeIndex& leftNode, PortIndex leftPortIndex, con
   setZValue(-1.0);
 
   // initialize the end points
-  if (leftNode.isValid()) {
+  if (leftNode.isValid())
+  {
     auto ngo = _scene.nodeGraphicsObject(leftNode);
     Q_ASSERT(ngo != nullptr);
 
     geometry().moveEndPoint(PortType::Out,  ngo->geometry().portScenePosition(leftPortIndex, PortType::Out, ngo->sceneTransform()));
 
   }
-  if (rightNode.isValid()) {
+  if (rightNode.isValid())
+  {
     auto ngo = _scene.nodeGraphicsObject(rightNode);
     Q_ASSERT(ngo != nullptr);
 
@@ -105,24 +113,26 @@ setGeometryChanged()
 
 ConnectionID
 ConnectionGraphicsObject::
-id() const {
+id() const
+{
   ConnectionID ret;
-  
+
   ret.lNodeID = _leftNode.id();
   ret.rNodeID = _rightNode.id();
   ret.lPortID = _leftPortIndex;
   ret.rPortID = _rightPortIndex;
-  
+
   return ret;
 }
 
 NodeDataType
-ConnectionGraphicsObject::dataType(PortType ty) const 
+ConnectionGraphicsObject::
+dataType(PortType ty) const
 {
   // get a valid node
   auto n = node(ty);
   Q_ASSERT(n.isValid());
-  
+
   return _scene.model()->nodePortDataType(n, portIndex(ty), ty);
 }
 
@@ -140,7 +150,7 @@ move()
       auto const &nodeGeom = nodeGraphics.geometry();
 
       QPointF scenePos =
-      nodeGeom.portScenePosition(portIndex(portType), 
+        nodeGeom.portScenePosition(portIndex(portType),
                                    portType,
                                    nodeGraphics.sceneTransform());
 
@@ -158,7 +168,9 @@ move()
 
 }
 
-void ConnectionGraphicsObject::lock(bool locked)
+void
+ConnectionGraphicsObject::
+lock(bool locked)
 {
   setFlag(QGraphicsItem::ItemIsMovable, !locked);
   setFlag(QGraphicsItem::ItemIsFocusable, !locked);
@@ -168,7 +180,7 @@ void ConnectionGraphicsObject::lock(bool locked)
 
 void
 ConnectionGraphicsObject::
-paint(QPainter* painter,
+paint(QPainter*                       painter,
       QStyleOptionGraphicsItem const* option,
       QWidget*)
 {
@@ -235,19 +247,22 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
   auto node = locateNodeAt(event->scenePos(), _scene,
                            _scene.views()[0]->transform());
 
-  if (!node) {
-    
-    if (state().requiresPort()) {
+  if (!node)
+  {
+
+    if (state().requiresPort())
+    {
       Q_ASSERT(this == _scene._temporaryConn);
       // remove this from the scene
       delete _scene._temporaryConn;
       _scene._temporaryConn = nullptr;
     }
-    
+
     return;
   }
 
-  if (!state().requiresPort()) {
+  if (!state().requiresPort())
+  {
     return;
   }
 
@@ -278,9 +293,9 @@ hoverEnterEvent(QGraphicsSceneHoverEvent* event)
   geometry().setHovered(true);
 
   update();
-  
+
   flowScene().model()->connectionHovered(_leftNode, _leftPortIndex, _rightNode, _rightPortIndex, event->screenPos(), true);
-  
+
   event->accept();
 }
 
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -258,7 +258,7 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
       Q_ASSERT(this == _scene._temporaryConn);
       // remove this from the scene
       _scene._temporaryConn = nullptr;
-      delete this;
+      deleteLater();
     }
 
     return;
@@ -275,16 +275,14 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
   {
     node->resetReactionToConnection();
     Q_ASSERT(this == _scene._temporaryConn);
-
-    delete _scene._temporaryConn;
     _scene._temporaryConn = nullptr;
+    deleteLater();
   }
   else if (state().requiresPort())
   {
     Q_ASSERT(this == _scene._temporaryConn);
-
-    delete _scene._temporaryConn;
     _scene._temporaryConn = nullptr;
+    deleteLater();
   }
 }
 
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -25,6 +25,7 @@ using QtNodes::ConnectionID;
 using QtNodes::FlowScene;
 using QtNodes::NodeDataType;
 using QtNodes::NodeIndex;
+using QtNodes::PortIndex;
 
 ConnectionGraphicsObject::
 ConnectionGraphicsObject(NodeIndex const& leftNode,
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -8,7 +8,6 @@
 
 #include "FlowScene.hpp"
 
-#include "Connection.hpp"
 #include "ConnectionGeometry.hpp"
 #include "ConnectionPainter.hpp"
 #include "ConnectionState.hpp"
@@ -22,13 +21,18 @@
 
 using QtNodes::ConnectionGraphicsObject;
 using QtNodes::Connection;
+using QtNodes::ConnectionID;
 using QtNodes::FlowScene;
+using QtNodes::NodeDataType;
 
 ConnectionGraphicsObject::
-ConnectionGraphicsObject(FlowScene &scene,
-                         Connection &connection)
-  : _scene(scene)
-  , _connection(connection)
+ConnectionGraphicsObject(const NodeIndex& leftNode, PortIndex leftPortIndex, const NodeIndex& rightNode, PortIndex rightPortIndex, FlowScene& scene)
+  : _state(leftNode.isValid() ? (rightNode.isValid() ? PortType::None : PortType::In) : PortType::Out)
+  , _leftNode{leftNode}
+  , _rightNode{rightNode}
+  , _leftPortIndex{leftPortIndex}
+  , _rightPortIndex{rightPortIndex}
+  , _scene{scene}
 {
   _scene.addItem(this);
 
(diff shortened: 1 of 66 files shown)


paceholder:
Oh, wow. I need to checkout it tomorrow and have a closer look. Impressive! Thanks

ypnos:
This looks very interesting. I already have an internal graph representation in my project (nodes, matching input/output types, etc.). As your PR doesn't change how the examples work, could you give a hint how I would profit when integrating my representation with your modelview approach?

With current master I need to write wrapper objects derived from `NodeDataModel`. Would that change?

russelltg:
Yes it would. This PR swings the storage of nodes from an internal strctuture to a subclass of 
 `FlowSceneModel`. `FlowSceneModel` is inspired by `QAbstractModel` and like all model-view classes in qt, it has a default model implementation for simplistic purposes (`QAbstrractListModel` has it's default implementtion used by `QListWidget` and uses `QListWidgetItem`s as elements). Similarly, `DataFlowWidget` uses a default model (defined in `DataFlowWidget.cpp`) that uses `Node`, `Connection`, and `NodeDataModel` instances as elements).

ypnos:
Ok, this looks quite awesome. I need to only write one class that does all the interfacing with my graph.

In your PR you have a consistent typo, it should be 'category' instead of 'catergory'.

russelltg:
Thanks much, I'll fix :)

ypnos:
@paceholder Would it be possible for you to give rough indication whether you are leaning towards accepting this PR, maybe in modified form, or not?

paceholder:
Hi guys,

sorry, busy times at work and at home -- couldn't get to the code.
The model-view approach is definitely something I'd like to have in the library. It will be merged as it is or with some corrections.

My plan is to mark the current state as version 1.0 and then continue with 2.x after merging this pull-request.

Best,

Dmitry

paceholder:
@russelltg your branch is not compilable,

ypnos:
`FlowSceneModel` is lacking a virtual destructor.

ypnos:
Thanks for the updates! There is a problem with drawing nodes in `NodePainter`. The individual node style (`model.nodeStyle(graphicsObject.index())`, see `drawNodeRect()`) should be used in places where currently `StyleCollection` is asked:

1. drawModelName
2. drawEntryLabels
3. drawValidationRect
4. drawConnectionPoints, drawFilledConnectionPoints

russelltg:
The only thing you've suggested that I haven't added is your number 6: 

> When I want to center a node under the cursor, I need to know the node's size. How do I get it under the new model? In DataFlowScene, you declare methods like getNodeSize(), getNodePosition(), but they are never defined.

russelltg:
I just realized I left most of the functions in `DataFlowScene` unimplemented and the signals never fired. Fixing rn.

russelltg:
That should be much better. Tell me what ya'll think!

russelltg:
@paceholder whats the status of this?

paceholder:
Sorry, we are awaiting a baby and I can't really work on side problems these days.

Regarding the code. This is a quite disruptive change. This would break the workflow of many users and bring lots of understanding problems. I can't merge it until I understand everything because I would have to support it further. I do not know how to handle it in a faster an most efficient way.

russelltg:
That's fair, no hurry. People who want to test out the changes can always
check our my fork.

Congrats on the baby!

src/ConnectionGraphicsObject.cpp
```diff
@@ -212,17 +241,39 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
 
   auto node = locateNodeAt(event->scenePos(), _scene,
                            _scene.views()[0]->transform());
+  
+  if (!node) {
+    
+    if (state().requiresPort()) {
```
ypnos: This is totally broken. First, it includes a heap use after free. Second, it is an object deleting itself under disguise. It is (apart from use after free) effectively the same as:
```c++
if (state().requiresPort()) {
    _scene._temporaryConn = nullptr;
    delete this;
    return;
}
```

Is this the desired intent?

src/ConnectionGraphicsObject.cpp
```diff
@@ -212,17 +241,39 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
...
+    return;
+  }
 
-  NodeConnectionInteraction interaction(*node, _connection, _scene);
+  NodeConnectionInteraction interaction(node->index(), *this);
 
   if (node && interaction.tryConnect())
   {
     node->resetReactionToConnection();
+    Q_ASSERT(this == _scene._temporaryConn);
```
ypnos: See my comment above.

src/ConnectionGraphicsObject.cpp
```diff
@@ -212,17 +241,39 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
...
+
+    delete _scene._temporaryConn;
+    _scene._temporaryConn = nullptr;
   }
-  else if (_connection.connectionState().requiresPort())
+  else if (state().requiresPort())
```
ypnos: Can this occur? In line 257 we return early for the opposite case. Did the state change in between? Also, see my comment above.

src/ConnectionGraphicsObject.cpp
russelltg: Yeah, that's exactly what I want, and i've fixed it to be exactly the code you've written and removed the use after free.

the destructor removes all references to this object and is safe to `delete this`. I just ran it under AddressSanitizer and once the use after free was fixed, everything worked as expected. 

Thanks for the catch!

ypnos:
You also might want to have a look at this commit:
https://github.com/ypnos/nodeeditor/commit/2ff9f39c21866958a2c8820b40ed97cc065220c3

russelltg:
Aah that's a good idea, will integrate!

ypnos:
A minor thing but have a look: https://github.com/ypnos/nodeeditor/commit/2ab945dd4885b2007cf7e837fe0156283fd72e5e

russelltg:
What's the need for those to be a part of the public API? They weren't a part of it in 1.0

ypnos:
Actually, the Connection and Node classes were part of the public API. Now that they are merged into the GraphicsObjects, they are lost to the public API.

The specific use-case is currently unknown for me.

russelltg:
Aaah yes of course, makes sense.

I'll pick that committ

wasawi:
Sorry for my confusion, I think i still don't fully understand the inner workings of this.

I have some questions:
- Does this design allow multiple input pins (of different data types) on a single module? Let's say i want a module to do something with a given image, string, and four floats. Would this design allow it?
- Is it possible to have a working graph, propagating data and computing results without actually having to draw the graph itself?

Thank you for any tips!
J

russelltg:
- yes it does. It functionally splits the rendering library and what it's actually representing ok the inside, so you can render anything (including your use case)
- not as of now, but this change makes it a pretty trivial change. I'll look into adding that into this PR with an example.

paceholder:
@russelltg  Honestly, I find this pr so disruptive that I can't figure out anymore how everything works. I'd prefer to implement this approach more gradually. Of course I'll take your changes into account

russelltg:
I don't know a way to implement it progressively, as it changes at the core
how it works. But I understand.

kollya:
@paceholder, I agree with Russel - it changes at the core, can't see how to implement similar stuff less painful. These changes make pretty good sense, waiting for a long time for them to be reviwed/merged by you!

russelltg:
Yep. I'm still willing to rebase and get this all working--I think it would be a big addition to the libary.

It would also greatly help those who want headless execution as it adds a split between graph and rendering.

russelltg:
@kollya The most what? Sorry I don't understand.

russelltg:
No it is pretty far behind.

I'll get it rebased in the coming days.

Quincunx271:
The travis tests are failing because you updated the Catch version to 2.3.0, but it also needs to be updated in the .travis.yml. Theres a line: `git checkout tags/v2.2.2`, which should be changed to `git checkout tags/v2.3.0` (or if there's a newer version, to that)

Quincunx271:
Actually, a version should probably be pinned in the .appveyor.yml too. It seems I forgot to do that when I added the tests

src/DataFlowScene.cpp
```diff
@@ -0,0 +1,709 @@
...
+
+Node&
+DataFlowScene::
+restoreNode(QJsonObject const& nodeJson)
+{
+  QString modelName = nodeJson["model"].toObject()["name"].toString();
+
+  auto uid = QUuid(nodeJson["id"].toString());
+ 
+  _dataFlowModel->addNode(modelName, point, uid);
```
Quincunx271: What is `point` here? MinGW [errors out here](https://ci.appveyor.com/project/paceholder/nodeeditor/build/1.0.530/job/c26y36xc0n5kracv#L210); it can't tell what variable `point` is supposed to be

src/DataFlowScene.cpp
```diff
@@ -0,0 +1,709 @@
...
+
+  auto& node = *_dataFlowModel->_nodes[uid];
+  node->restore(nodeJson);
```
Quincunx271: Should this be `node.restore(nodeJson)`? `node` is a reference here, not a pointer.

russelltg:
Everything should be compiling now, except the tests aren't passing. I could use a bit of mentorship on debugging this, I haven't dealt with this sort of tests before.

Quincunx271:
Doesn't seem to be compiling. The [errors](https://travis-ci.org/paceholder/nodeeditor/jobs/419254710#L1109) say something about accessing a private member of `ConnectionID`

paceholder:
@russelltg I fixed that in my branch `modelview`. I don't know how to make a pull-request into a pull-request

russelltg:
@paceholder I can just cherry pick the commit. 

PS: you can go to chigraph/nodeeditor and check new PR and select the right branches :)

paceholder:
> PS: you can go to chigraph/nodeeditor and check new PR and select the right branches :)

well... no. It is a fork of my repository, GitHub is having troubles with doing a PR backwards. At least I did not figure out how to select the branches (paceholder -> chigraph)

image

russelltg:
https://github.com/chigraph/nodeeditor/compare/modelview...paceholder:modelview

russelltg:
I added a basic example for running it headless, let me know what you think.

russelltg:
Right now, `TestNodeGraphicsObject` uses private APIs (`NodeGraphicsObject`), so it won't link on windows as it's not being exported. 

What's the right solution here?

Quincunx271:
The best solution would be to not use the mouse at all; that would remove the need for the `geometry` function. However, that would require refactoring `NodeGraphicsObject::mousePressEvent()` so that the test code could attempt to create a connection without the mouse. (It is currently possible to create connections without the mouse, but the code that had the bug was inside the `mousePressEvent` function).

kollya:
msvs 2017 x64, latest Qt 5.11.2 - all examples crash on:
- add a node
- add another node
- click the 1st node's output and drag the connection line to the 2nd node's input
- releasing the mouse above the 2nd node's input
- crash

kollya:
Looks like a fix for the above is in this commit (but requires also other changes to apply it as it is):
https://github.com/paceholder/nodeeditor/commit/d4b83952c62b41ef1ad61dbec27639f409b9a6b8#diff-dac991e3dc5e8f4da298fd5ee7b04ca8

kollya:
**russelltg**,

Could you please clarify what was your motivation for making input ports accepting only one connection? Why not to make input ports using ConnectionPolicy - as output ports use?

P.S. Going further, we could have more detailed restriction: 1, many or exactly N. Even further - a method that returns if a connection should be accepted, depending on the current number of connections, nodetype of the connection's other side, etc. But I don't have real use cases for that right now.

russelltg:
I'm not sure what you mean--all ports can define their own ConnectionPolicy

https://github.com/chigraph/nodeeditor/blob/bf7a0fc3bdd387a9c483b65f3d01f43032dd8cfa/include/nodes/internal/FlowSceneModel.hpp#L122-L124

Possibly you mean in `DataFlowModel`, which is a implementation of `FlowSceneModel` specificly for data processing. It's possible that `portOutConnectionPolicy` should be returned entirely from `NodeDataModel`.

russelltg:
Also I cannot reproduce the bug on linux, I will try to repro on Windows soon.

Quincunx271:
In this case, that code is UB. It's okay to `delete this` if you are extremely careful: https://isocpp.org/wiki/faq/freestore-mgmt#delete-this . However, this case is not fine:

> You must be absolutely 100% positively sure that the rest of your member function (after the delete this line) doesn’t touch any piece of this object (**including calling any other member functions or touching any data members**). This includes code that will run in destructors for any objects allocated on the stack that are still alive.

<sup>emphasis added</sup>

In this case, the second line accesses the `_scene` data member. If you absolutely need to access `_scene` afterwards, you'd have to store it as a stack variable and access this local variable.

----

If the code could be reworked to not have to `delete this`, it's much easier to verify its correctness.

Quincunx271:
> Also I cannot reproduce the bug on linux, I will try to repro on Windows soon.

For reproing things like this on linux, try compiling and linking with `-fsanitize=address`, `-fsanitize=undefined`, or `-fsanitize=memory`. With cmake, you can set the `CXXFLAGS` and `LDFLAGS` env variables prior to invoking cmake:

```
$ CXXFLAGS=-fsanitize=address LDFLAGS=-fsanitize=address cmake ..
```

russelltg:
You're definitely right. Should be fixed.

krsvojte:
I had the same issues as https://github.com/paceholder/nodeeditor/pull/117#issuecomment-430542057 (same system config) however the commit https://github.com/paceholder/nodeeditor/pull/117/commits/c235e3940f87fffe8d866fd98174571a7ac8b5a1 didn't fix the issue.

After some experimentation, I found two bugs:

1. `ConnectionGraphicsObject::mouseReleaseEvent `has two more deletes and replacing them with
`_scene._temporaryConn = nullptr;
	delete this;` fixes that crash.

2. Another crash kept happening in QGraphicsScene BSP tree. It would happen intermittently, other times there would be a heap corruption upon exiting the application. There seemed to be an invalid pointer to QGraphicsItemPrivate left in the BSP tree. It can be fixed by adding this line:
`this->removeItem(&cgo);`
before 
`delete &cgo;`
in `FlowScene::connectionRemoved`. This makes sure the removed connection graphics item is removed from the scene (and ownership of the pointer is transferred back to caller and it's safe to delete).

kollya:
> I had the same issues as [#117 (comment)](https://github.com/paceholder/nodeeditor/pull/117#issuecomment-430542057) (same system config) however the commit [c235e39](https://github.com/paceholder/nodeeditor/commit/c235e3940f87fffe8d866fd98174571a7ac8b5a1) didn't fix the issue.
> 
> After some experimentation, I found two bugs:
> 
> 1. `ConnectionGraphicsObject::mouseReleaseEvent `has two more deletes and replacing them with
>    `_scene._temporaryConn = nullptr; delete this;` fixes that crash.
> 2. Another crash kept happening in QGraphicsScene BSP tree. It would happen intermittently, other times there would be a heap corruption upon exiting the application. There seemed to be an invalid pointer to QGraphicsItemPrivate left in the BSP tree. It can be fixed by adding this line:
>    `this->removeItem(&cgo);`
>    before
>    `delete &cgo;`
>    in `FlowScene::connectionRemoved`. This makes sure the removed connection graphics item is removed from the scene (and ownership of the pointer is transferred back to caller and it's safe to delete).

regarding the crash in ConnectionGraphicsObject::mouseReleaseEvent:

```
    Q_ASSERT(this == _scene._temporaryConn);

    delete _scene._temporaryConn;
    _scene._temporaryConn = nullptr;
    delete this;
  }
```

1. delete _scene._temporaryConn; // this line delete *this*, because _scene._temporaryConn == this
2. _scene._temporaryConn = nullptr; // crash here. *this* already deleted;

kollya:
I also had crashes (inside QGraphicsSceneBspTree methods) , scenario is:
1. add two nodes
2. drag a connection from node1's output towards node2
3. release the mouse above node2 area (not above it's input, but the body)
4. if temp connection disappeared and everything is fine, go to step 2.
5. crash

Call stack etc:
https://gyazo.com/1a93a6b36e14259577c6cbf883a39b0a

I suppose that fix would be:
instead of deleting QObject with 'delete this' inside it's event handler we should use QObject::deleteLater()

kollya:
pull request for the fix described above: https://github.com/chigraph/nodeeditor/pull/3/

aol-nnov:
total overconcentrated awesomeness over here! so glad I've stumbled upon this project yesterday!

@paceholder thank you for such an awesome contribution, you rock!!

@russelltg  model/view separation is a perfect idea and I'd like to dive into this pull request and, eventually, help making this feature to land into upstream.

Would you mind resolving merge conflicts meanwhile?

Thank y'all, guys! Fantastic work!

Daguerreo:
@russelltg great work!
@paceholder there are news about this update?

paceholder:
Hi @Daguerreo,

sorry, I have no time to work on a project now. Feel free to clone this branch and use in your project.

Best,

Dmitry

Daguerreo:
@russelltg how should be fixed #198 in this branch?

russelltg:
I'm pretty sure cycles are allowed in this branch, it depends on the model you use though. It's the model's fault (specifically DataFlowModel)

tatatupi:
Hi everyone. 

I have been using nodeeditor for a scientific software under development for a few months. I am considering changing the branch for the modelview one, since our code is based in the model view philosophy.

However, I would like to hear from you guys, @paceholder and @russelltg, if you advise against it for some reason. I noticed that the master branch from paceholder is still active, with commits from July, whereas the lattest commit from chigraph's modelview fork is from Nov-2018.

I have a feeling that the modelview branch is not going to be merged into master anymore, and they are somehow being supported in parallel.

Daguerreo:
> Hi everyone.
> 
> I have been using nodeeditor for a scientific software under development for a few months. I am considering changing the branch for the modelview one, since our code is based on the model view philosophy.
> 
> However, I would like to hear from you guys, @paceholder and @russelltg, if you advise against it for some reason. I noticed that the master branch from paceholder is still active, with commits from July, whereas the lattest commit from chigraph's modelview fork is from Nov-2018.
> 
> I have a feeling that the modelview branch is not going to be merged into master anymore, and they are somehow being supported in parallel.

He has no intention of merging it. I have no idea of the reason he's no answering about it though.

paceholder:
Hi @Daguerreo,

I did not really change much in the master branch during the last year. I think you could go ahead and use @russelltg 's branch.

@russelltg with all my respect to your large and impressive work I will decline this pull request in order not to confuse people and to not force them to wait for something.
Here is my reasoning:
1. The full request is quite complex, I spotted many confusing places and maybe several potential bugs.
2. It has no code documentation and no unit-tests.
3. This PR is a breaking change making the project hardly maintainable for me in contrast to my original motivation of making "a proof of concept"
4. The model-view approach is still the correct one, this that I should have chosen from the beginning. Thus I will keep it in a separate branch called *chigraph-modelview*
5. I will come back to this idea later. I have quite a huge side project in mind and this model will be just a part of it. I am working in my spare time (several hours a week) implementing my ideas after my main job. 

I hope you understand my motivation and will still use this lib for your own benefits.

Daguerreo:
@paceholder don't worry, we all understand and there is no shame in saying "thanks but no". A negative answer is always better than not answering though :)

Btw, it was @tatatupi question, not mine :D

russelltg:
Yeah---it is quite behind at this point, and I don't have much time to pursue it.

And yes, there are likely bugs as I've had to rebase it a few times as it's old. Sad to see nonetheless.

tatatupi:
OK. Thank's everyone. We had a couple of issues with the modelview branch. And since @russelltg has mentioned his lack of time, we will go with @paceholder master branch!

PR #118, opened 2017-09-25
Allow setting node style from file

ypnos:
Let an application ship their global style in a file, e.g.:
```c++
const QString styleFile = ":NodeEditorStyle.json";
QtNodes::ConnectionStyle::setConnectionStyleFromFile(styleFile);
QtNodes::FlowViewStyle::setStyleFromFile(styleFile);
QtNodes::NodeStyle::setNodeStyleFromFile(styleFile);
```

Also remove redundant constructor to avoid interface bloat.

### DIFF
--- a/src/ConnectionStyle.cpp
+++ b/src/ConnectionStyle.cpp
@@ -28,19 +28,23 @@ ConnectionStyle()
 }
 
 
+void
 ConnectionStyle::
-ConnectionStyle(QString jsonText)
+setConnectionStyle(QString jsonText)
 {
-  loadJsonFile(":DefaultStyle.json");
-  loadJsonText(jsonText);
+  ConnectionStyle style;
+  style.loadJsonText(jsonText);
+
+  StyleCollection::setConnectionStyle(style);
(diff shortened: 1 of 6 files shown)


paceholder:
My old implementation wasn't the best one and this one does not improve the situation.

I do not see why you couldn't load styles from file without this pull-request.
The style classes accept QString -- a json style -- and this was done on purpose. You could get this text from whatever you want. It is a very general interface.

The function  `NodeStyle::loadJsonFile` was implemented private on purpose as well. It hides the fact that we store the default style in a resource file. This has nothing to do with the external world.

If you look at the example `styles` you'd see that we construct a text-style at the beginning of the `main.cpp`. You could put your file-stuff there.

ypnos:
I agree the styles interfaces and code structure is a bit clumsy. My intend was to stay minimally invasive.

I don't understand why you implement functionality and then deliberately hide it from possible users. After all, reading a style from a resource is not an obscure use case.

Next time I will ask before doing the work.

PR #135, opened 2017-12-07
Removed the need for clone()

facontidavide:
Hi,

I suggest to change the DataModelRegistry to store a function pointer of type 

        std::function<RegistryItemPtr()>

instead of a "prototype" instance of RegistryItemPtr that is used to call **clone()**.

The advantage is twofold:

- No need to have a pure visrtual function clone() anymore
- It was impossible to have derived classes of NodeDataModel which need arguments in the constructor. Now it is possible to use std::bind or capturing lambdas to deal with this use case.

For instance:

``` c++

class MyDataModel : public NodeDataModel
{
  Q_OBJECT

public:
  MyDataModel(QString name) : _name(name) {}

QString name() const override  { return QString("MyDataModel:") + _name; }

// More stuff
private:
   QString _name;
}

// code where you register
{
  auto ret = std::make_shared<DataModelRegistry>();
  ret->registerModel<MyDataModel>( [](){ return std::make_unique<MyDataModel>("NameA") ;  );
  ret->registerModel<MyDataModel>( [](){ return std::make_unique<MyDataModel>("NameB") ;  );
}

```

### DIFF
--- a/src/DataModelRegistry.cpp
+++ b/src/DataModelRegistry.cpp
@@ -10,11 +10,11 @@ std::unique_ptr<NodeDataModel>
 DataModelRegistry::
 create(QString const &modelName)
 {
-  auto it = _registeredModels.find(modelName);
+  auto it = _registeredModelCreators.find(modelName);
 
-  if (it != _registeredModels.end())
+  if (it != _registeredModelCreators.end())
   {
-    return it->second->clone();
+    return it->second();
   }
 
   return nullptr;
@@ -25,7 +25,7 @@ DataModelRegistry::RegisteredModelsMap const &
 DataModelRegistry::
 registeredModels() const
 {
-  return _registeredModels;
+  return _registeredModelCreators;
 }
 
 
@@ -54,7 +54,7 @@ getTypeConverter(QString const &sourceTypeID, QString const &destTypeID) const
 
   if (converter != _registeredTypeConverters.end())
   {
-    return converter->second->Model->clone();
+    return converter->second->ModelCloner();
   }
   return nullptr;
 }
--- a/src/DataModelRegistry.hpp
+++ b/src/DataModelRegistry.hpp
@@ -3,6 +3,7 @@
 #include <unordered_map>
 #include <set>
 #include <memory>
+#include <functional>
 
 #include <QtCore/QString>
 
@@ -20,15 +21,16 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 public:
 
   using RegistryItemPtr             = std::unique_ptr<NodeDataModel>;
-  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemPtr>;
+  using RegistryItemCloner          = std::function<RegistryItemPtr()>;
+  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemCloner>;
   using RegisteredModelsCategoryMap = std::unordered_map<QString, QString>;
   using CategoriesSet               = std::set<QString>;
 
   struct TypeConverterItem
   {
-    RegistryItemPtr Model{};
-    NodeDataType    SourceType{};
-    NodeDataType    DestinationType{};
+    RegistryItemCloner ModelCloner{};
+    NodeDataType       SourceType{};
+    NodeDataType       DestinationType{};
   };
 
   using ConvertingTypesPair = std::pair<QString, QString>; //Source type ID, Destination type ID in this order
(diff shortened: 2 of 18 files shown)


Quincunx271:
I agree with this idea, though I haven't looked through the PR itself. It makes the library easier to understand (`clone()` is a confusing name as it's only used for creating a new node).

Another benefit of this is that it allows for completely different data between the factory and the objects themselves. For example, in my code, this would allow me to convert a `shared_ptr` in every node into just a raw pointer to something stored in the factory object.

facontidavide:
Are you conceptually interested to this PR? Do you want me to update it to fix the CI issue?

paceholder:
Hi @facontidavide,

yes, I think this is the right way to go. Sorry I made a couple of other changes in the library and did not get to this one. 

It would be very kind of you to update the pull-request.

Thanks

Dmitry

src/DataModelRegistry.cpp
```diff
@@ -54,7 +54,7 @@ getTypeConverter(QString const &sourceTypeID, QString const &destTypeID) const
 
   if (converter != _registeredTypeConverters.end())
   {
-    return converter->second->Model->clone();
+    return converter->second->ModelCloner();
```
Quincunx271: I recommend using the name `ModelCreator` rather than `ModelCloner`. We are removing the `clone()` operation in favor of a factory function (there's `_registeredModelCreators`)

This is a member of `TypeConverterItem`, defined [here](https://github.com/paceholder/nodeeditor/pull/135/files#diff-3836c6057311bf852253b1fb88143a55R31)

src/DataModelRegistry.hpp
```diff
@@ -20,15 +21,16 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 public:
 
   using RegistryItemPtr             = std::unique_ptr<NodeDataModel>;
-  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemPtr>;
+  using RegistryItemCloner          = std::function<RegistryItemPtr()>;
```
Quincunx271: `using RegistryItemCreator`

src/DataModelRegistry.hpp
```diff
@@ -50,36 +52,36 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 
   template<typename ModelType, bool TypeConverter = false>
   void
-  registerModel(std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>(), QString const &category = "Nodes")
+  registerModel(RegistryItemCloner cloner = [](){ return std::make_unique<ModelType>(); },
```
Quincunx271: `registerModel(RegistryItemCreator creator = `

Lambdas with no arguments can be defined without the `()`s, so you could just do `[] { return std::make_unique<ModelType>(); }`, but that honestly doesn't really matter

src/DataModelRegistry.hpp
```diff
@@ -89,9 +91,9 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
   //Parameter order alias, so a category can be set without forcing to manually pass a model instance
   template<typename ModelType, bool TypeConverter = false>
   void
-  registerModel(QString const &category, std::unique_ptr<ModelType> uniqueModel = std::make_unique<ModelType>())
+  registerModel(QString const &category, RegistryItemCloner cloner = [](){ return std::make_unique<ModelType>();})
```
Quincunx271: `RegistryItemCreator creator`

src/DataModelRegistry.hpp
```diff
@@ -20,15 +21,16 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
...
+  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemCloner>;
   using RegisteredModelsCategoryMap = std::unordered_map<QString, QString>;
   using CategoriesSet               = std::set<QString>;
 
   struct TypeConverterItem
   {
-    RegistryItemPtr Model{};
-    NodeDataType    SourceType{};
-    NodeDataType    DestinationType{};
+    RegistryItemCloner ModelCloner{};
```
Quincunx271: `RegistryItemCreator ModelCreator`

src/DataModelRegistry.hpp
```diff
@@ -20,15 +21,16 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 public:
 
   using RegistryItemPtr             = std::unique_ptr<NodeDataModel>;
-  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemPtr>;
+  using RegistryItemCloner          = std::function<RegistryItemPtr()>;
+  using RegisteredModelsMap         = std::unordered_map<QString, RegistryItemCloner>;
```
Quincunx271: Maybe `RegisteredModelCreatorsMap`?

Quincunx271:
I thought about this a bit more, and I think it *might* be beneficial to use our own Factory type rather than `std::function`. The strongest benefit that I can see is that it makes it possible in the future for us to add things to this "factory", stuff that are static per `NodeDataModel` subclass.

This is similar to the class model in Java, where there's a single `Class` object that represents the type of all objects of that class. Or in C++, it'd be a lot like `static` member functions of a type.

This would greater flexibility in the future. Stuff like the name of the node type could potentially be stored in it.

facontidavide:
I move this to a different PR. @Quincunx271 , thanks for the suggestions.

Issue #127, opened 2017-10-24
NodeDataModel::portOutConnectionPolicy is called for input connections as well


Quincunx271:
`portOutConnectionPolicy` should definitely not be called for input ports, but that's exactly what happens.

This is the shortest code I could throw together to reproduce the bug:

```c++
#include <nodes/ConnectionStyle>
#include <nodes/DataModelRegistry>
#include <nodes/FlowScene>
#include <nodes/FlowView>
#include <nodes/NodeData>
#include <nodes/NodeDataModel>

#include <QtCore/QObject>
#include <QtWidgets/QApplication>
#include <QtWidgets/QLabel>
#include <QtWidgets/QMenuBar>
#include <QtWidgets/QVBoxLayout>

#include <iostream>

using QtNodes::DataModelRegistry;
using QtNodes::NodeDataModel;
using QtNodes::FlowScene;
using QtNodes::FlowView;
using QtNodes::ConnectionStyle;

namespace {
    class FooModel : public NodeDataModel {
    public:
        using PortType = QtNodes::PortType;
        using Data = QtNodes::NodeData;
        using DataType = QtNodes::NodeDataType;

        FooModel() = default;

        QString caption() const override { return "Foo"; }

        QString name() const override { return "Foo"; }

        std::unique_ptr<NodeDataModel> clone() const override { return std::make_unique<FooModel>(); }

        unsigned int nPorts(PortType type) const override
        {
            switch (type) {
            case PortType::In:
                return 2;
            case PortType::Out:
                return 1;
            }
            throw "error";
        }

        DataType dataType(PortType, int index) const override { return { "data", "type" }; }
        std::shared_ptr<Data> outData(int index) override { return nullptr; }
        void setInData(std::shared_ptr<Data>, int index) override {}

        QWidget* embeddedWidget() override { return nullptr; }

        QtNodes::NodeDataModel::ConnectionPolicy portOutConnectionPolicy(int index) const override
        {
            std::cout << "Index: " << index << '\n';
            return QtNodes::NodeDataModel::ConnectionPolicy::One;
        }
    };

    std::shared_ptr<DataModelRegistry> registerDataModels()
    {
        auto ret = std::make_shared<DataModelRegistry>();
        ret->registerModel<FooModel>();

        return ret;
    }
}

int main(int argc, char* argv[])
{
    QApplication app(argc, argv);

    QWidget mainWidget;

    QVBoxLayout* l = new QVBoxLayout(&mainWidget);

[...]

PR #140, opened 2018-02-26
Fix Issue #127

Quincunx271:
Fix the bug raised in issue #127


Quincunx271:
The travis build appears to be failing because of a missing Qt5 dependency; I don't know what would cause that. It's unrelated to the changes I made in this PR

Issue #141, opened 2018-02-27
Add CMake installs - Fully modernize CMakeLists.txt


Quincunx271:
[Relevant article on how to do CMake "right"](https://pabloariasal.github.io/2018/02/19/its-time-to-do-cmake-right/).

Currently when you build the library, you'd have to manually move the static library. If you set up the `install` commands correctly, it becomes possible for one to just run `make install`, then `find_package(nodeeditor)` will include the exported static library and headers.

This is not the easiest change, but it's doable. Some things it would require:

* Rewriting the top level CMakeLists.txt
* Reorganizing the `include/` directory; rather than simply forwarding headers from the `src/` directory, those `src/` headers should be moved to `include/` (implementation headers would remain in `src/`)

PR #142, opened 2018-02-28
Improve cmake (modernize, add install()s)

Quincunx271:
This PR addresses #141 

It has a notable change to the project structure: all the public headers, even if they were under `src/`, were moved under `include/`. This means that installing the entire `include/` directory doesn't miss any required headers. It might be possible to install the public headers from `src/`, but the added complexity is not worth it in my opinion (my best idea is to manually keep a list of all the headers which should be exported)

Some changes were very clear-cut changes to follow CMake best practices. Here are some of those:
* `add_definitions` -> `target_compile_definitions`
* `set(CMAKE_CXX_FLAGS ...)` -> `target_compile_options`
* `add_definitions` which were really trying to add compiler flags -> `target_compile_options`
* `set(CMAKE_CXX_STANDARD ...)` -> `target_compile_features`
* `set(CMAKE_EXPORT_COMPILE_COMMANDS "ON")` -> removed. It can be set at configure time with `cmake <...> -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`

There is one rather large change that some people question:
* `file(GLOB_RECURSE ...)` for obtaining source files -> manually listing in `add_library` (or `target_sources`).

The reason for manually listing is that you don't have to manually rerun `cmake` every time you add/remove a source file; the build system detects that _CMakeLists.txt_ has changed and will rerun it for you. If this particular change is not desired, it's not a big deal to revert it. In my own experience, it's not much effort to add a line for the file every time you make a new file.

Now, after installing the library with `make install`, a user can write `find_package(NodeEditor REQUIRED)` and in `target_link_libraries`, link with `NodeEditor::nodes`. The exact names for both are not too difficult to change, but they seemed like logical choices, with the project name being `NodeEditor` and the target being `nodes`.

The two commits have several inline github comments.


paceholder:
Hi,

thanks a lot for this great work. I'll look through and then merge it.

PR #144, opened 2018-03-11
Modify source to be compatible with Qt 5.5.1 (Visual Studio 2010)

iclosure:


Quincunx271:
This is a pretty substantial change. I have some comments

* In *examples/calculator/DecimalToIntegerModel.cpp* `std::make_shared<IntegerData>(int(_decimal->number()));`, I'm unsure why you need to cast to `int`, but I believe it would be better to make it a `static_cast`: `std::make_shared<IntegerData>(static_cast<int>(_decimal->number()));`
* In *examples/calculator/DivisionModel.hpp*, I don't know why you add parameter names to `portCaptionVisible` only to ignore them with `Q_UNUSED`; why not just leave the parameters unnamed?
* There are more little changes that I just don't understand in the examples, but it's really mostly nitpicking.

There are some other more questionable changes IMO:

* You change a lot of constructors `Foo() = default;` to `Foo();` with the constructor explicitly written out in a source file. There's no reason to do this. Similarly with destructors, there are a lot of `~Foo() = default;` replaced with `~Foo() {}`. The `= default;` pattern is considered more idiomatic C++
* Replacing public constructors written as `= delete;` with private constructors is harmful to error messages. For example, `Connection`'s copy ctor and copy assignment operator.
* I disagree with replacing the type aliases of the form `using Name = ...;` with `typedef ... Name;`
* I don't understand why you removed the default `TypeConverter = false` on `registerModel`; it just makes it harder to use the API
* You added QtCreator *.pro* files. This library has been using CMake rather than QtCreator projects. Maybe someone can change the *.pro* files into CMake project files.
* In *Export.hpp*, prefer type-aliases to `#define`s; the `#define UnorderedMap std::map` can easily break a user's code
* I don't understand using this weird `foreach` keyword instead of just a regular C++ range-based for loop.

Overall, I'm extremely uncomfortable with these changes. I think that supporting Qt 5.5.1 is a reasonable goal, but Visual Studio 2010 supports very few C++11 features ([here is some documentation I found](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015)). Trying to support VS 2010 seems like it's likely to make the code a lot worse. I think it's a reasonable requirement to require C++11

I don't mean to downplay the work you put in, but these changes will make the library unusable for me (won't compile), or at best will make it harder to work with.

paceholder:
As I understood, the whole bunch of changes was done to avoid using C++11. This is an absolute no go for this project. Please, update your compiler. It is 2018.

iclosure:
Ok! I understood,Thanks for your response

PR #145, opened 2018-03-12
Fix -fPIC/-fPIE travis

Quincunx271:
Fixes the bug where travis complained about needing to compile with `-fPIC`. The reason for this was because it was compiling with both `-fPIE -fPIC`, which seems to default to just `-fPIE`. You cannot link Position Independent Executable code into a shared library. It has to be Position Independent Code ([source](https://lists.debian.org/debian-devel/2016/05/msg00309.html)).

This extra `-fPIE` flag was added by `${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}`. As we are making a library, not an executable, we don't want this. The fix is to remove it altogether, as it's added automatically when the user links with the Qt libraries (whether directly or transitively through us)

Additionally, in *.travis.yml*, I clarified some comments, and added a command to make it easier to get a verbose build (`make VERBOSE=1`). Now, if the build fails, you can go to https://travis-ci.org/paceholder/nodeeditor/settings and add a `VERBOSE_BUILD` environment variable. When `VERBOSE_BUILD` is set to `ON`, it will act as if you had `make VERBOSE=1`. When `VERBOSE_BUILD` is unset or set to `OFF`, it will act normally (non-verbosely). You can then trigger a rebuild (no commit needed) to get the verbose information (or you can just leave `VERBOSE_BUILD=ON` and always have verbose information)


paceholder:
Thanks a lot. Interesting, previously we added the flags from `${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}` as well.  And it is somehow gets compiled and linked on my laptop (Linux, Qt 5.10, gcc 7.3)

Quincunx271:
@paceholder It compiles and links just fine on my Ubuntu 16.04 system as well. On my system, `${Qt5Widgets_EXECUTABLE_COMPILE_FLAGS}` expands to `-fPIC` rather than `-fPIE`. Likely has to do with the version of Qt, perhaps it was a bug in Qt. Either way, that was the problem.

PR #150, opened 2018-04-04
Implement Converters separately from DataModels

paceholder:
It is not completely finished yet but demonstrates the idea.

The notion of data converters is separated from DataModels.

A converter is anything callable (lamda or functor) wrapped into `std::function` with the signature `std::shared_ptr<NodeDataModel>(std::shared_ptr<NodeDataModel>)`.

All converters are stored in the mapping `(data_id1, data_id2) -> std::function<Converter>`.

TODO:
- json save/restore
- paint some icon on top of the "converting" connection
- use color gradient for "converting" connections when they are painted according to data types.


paceholder:
Hi @Quincunx271 

I did not find a way to add you to rewiewers.
image

There are also some clang compilation issues to be fixed

Dmitry

include/nodes/internal/DataModelRegistry.hpp
```diff
@@ -7,9 +7,21 @@
...
+namespace std {
+
+  inline
+  bool operator<(QtNodes::NodeDataType const & d1,
+                 QtNodes::NodeDataType const & d2)
+  {
+    return d1.id < d2.id;
+  }
+}
+
```
Quincunx271: This may be the cause of [this compilation error](https://travis-ci.org/paceholder/nodeeditor/jobs/362367315#L1053):

```
/usr/bin/../lib/gcc/x86_64-linux-gnu/5.4.1/../../../../include/c++/5.4.1/bits/stl_pair.h:221:24: error: 
      call to function 'operator<' that is neither visible in the template
      definition nor found by argument-dependent lookup
    { return __x.first < __y.first
                       ^
```

Firstly, **it's not allowed to add anything in `namespace std`**, unless it's explicitly stated that you can (e.g. specializations of `std::hash`).

Second, this is unnecessary. You can add the `operator<` to the `QtNodes` namespace, and [Argument Dependent Lookup (ADL)](http://en.cppreference.com/w/cpp/language/adl) will find it. In other words, just add this `operator<` following the declaration of `NodeDataType`, then `dataType1 < dataType2` will find the correct function.

include/nodes/internal/Converter.hpp
```diff
@@ -0,0 +1,25 @@
...
+#include <tuple>
+
+namespace QtNodes
+{
+
+using SharedNodeData = std::shared_ptr<NodeData>;
+
+// a function taking in NodeData and returning NodeData
+using Converter =
+  std::function<SharedNodeData(SharedNodeData)>;
```
Quincunx271: I think that `TypeConverter` might be a better name. My first thought when I see `Converter` is: "What is this converting?" Maybe `DataConverter`?

include/nodes/internal/DataModelRegistry.hpp
paceholder: > Firstly, it's not allowed to add anything in namespace std

I used a lot `std::unordered_map` at work and blindly wrote the `operator<()` as if I were implementing the `hash` function. I'll fix that.

paceholder:
> Note that you are making a breaking change. 

I know. But it is impossible to keep the current design infinitely. I will update the version to `2.0` with this change.

There might be some further disrupting changes:
- Consider ditching the `QString`-based data type id and look closer at [`c++ typeid`](http://en.cppreference.com/w/cpp/language/typeid)
- Rework library into a general-purpose node-editor. This might help in cases where user needs two connections merging in out input-port. Data-flow would work on top of it as one of possible use cases.
- Allowing the use of external graph representations.

CMakeLists.txt
```diff
@@ -80,7 +80,7 @@ target_compile_definitions(nodes
 
 target_compile_options(nodes
   PRIVATE
-    $<$<CXX_COMPILER_ID:MSVC>:/W4 /EHsc>
+    $<$<CXX_COMPILER_ID:MSVC>:/W4 /wd4127 /EHsc>
```
paceholder: Tons of warnings from Qt code

paceholder:
I haven't finished with the colors yet. I pointed it out in the very first comment -- I'll add connection-color gradients, some fancy icon to tag the "converting" connections and JSON store\restore

Quincunx271:
Whoops, I didn't see that.

paceholder:
image

Quincunx271:
I wonder if it should be a gradient or just a color change at the point of the conversion symbol. I'm just noticing how the connection appears more cyan than red to me there, and wondering if there are color combinations for which it would become hard to notice. I bet it looks better for some color combinations than others.

paceholder:
Merged manually and pushed to master

PR #159, opened 2018-04-27
Fixes crash when loading previously saved file with connections

### LINKED ISSUE #158: Crash when loading a json file
I've tried out the calculator example. Everything compiled without problems. I started the calculator example and made a sketch and saved it but when I wanted to load the same file the program stopped working.

I could trace the error to this function: `NodeState::setConnection`

It consists of these two lines:
`auto &connections = getEntries(portType);
connections[portIndex].insert(std::make_pair(connection.id(), &connection));`

Somehow the address of `connections` is always 0x20, which is not a valid address, so the second line results in a segmentation fault.

This only happens when loading the connections from a JSON file.

DanielBorn:
This fixes #158

### DIFF
--- a/src/Connection.cpp
+++ b/src/Connection.cpp
@@ -87,10 +87,10 @@ save() const
 
   if (_inNode && _outNode)
   {
-    connectionJson["in_uid"] = _inNode->id().toString();
+    connectionJson["in_id"] = _inNode->id().toString();
     connectionJson["in_index"] = _inPortIndex;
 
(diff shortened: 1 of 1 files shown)


paceholder:
Great, thanks a lot!
I see now what happened. I renamed the internal class member `Connection::_id` to `Connection::_uid` by blindly replacing all the entries in the source file and broke this function.
The original name `_id` was conflicting with the `moc`-generated code.

Dmitry

PR #162, opened 2018-04-30
Move implementation of std::make_unique to QtNodes::detail::make_unique

Quincunx271:
Technically, it's not allowed to add things to `namespace std`. It may be true that opening it up to add a `make_unique` implementation isn't going to break on any of the compilers+standard libraries at this time, but it's [technically undefined behavior](http://en.cppreference.com/w/cpp/language/extending_std). In theory, compilers could do something like make it a compile error.

The only way to be technically compliant is to drop support for environments without `std::make_unique` or to have your own `make_unique` function in your own namespace.

This PR does the latter: it moves the implementation of `make_unique` into `QtNodes::detail` and turns calls from `std::make_unique` into `detail::make_unique`


facontidavide:
I originally introduced the make_unique for older compilers.

I agree with you, better to play it safe with a different namespace. +1

paceholder:
Hi Justin,

could you please explain

> Perhaps we should add an appveyor check that uses a version of the MS STL without std::make_unique.

Thanks

Quincunx271:
@paceholder Appveyor currently checks MinGW Makefiles and Visual Studio 2015 v14. From wikipedia, [we can see that](https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering) `_MSC_VER` is `1900`. Our implementation of `make_unique` is made active when `_MSC_VER < 1800`, which is VS 2012 and earlier. My suggestion was to add a build on Appveyor which uses VS 2012, so that there wouldn't be a `std::make_unique`. This would mean that anywhere that a `std::make_unique` was accidentally used rather than `detail::make_unique`, the build would fail.

Quincunx271:
@paceholder Here's another idea: we could add a check in the travis build to ensure there is no use of `std::make_unique` outside of the one file. It would be undesirable for this check to stop the build when it fails, though.

For example, this command would show any accidental uses, and its error code would only be 0 if there are no accidental uses:

```
! grep -nre 'std::make_unique' --exclude='memory.hpp' include src
```

Issue #147, opened 2018-03-27
Allow arbitrary ordering of nodes in the context menu


Quincunx271:
The ordering of nodes inside the DataModelRegistry is currently pretty arbitrary. When you open the context menu, the nodes are ordered by Category (sorted, as it uses `std::set`), but within each category, there is no ordering. If I understand the code correctly, this is because `RegisteredModelsCategoryMap` is a `std::unordered_map`.

I think it's beneficial to allow this ordering to be fully specified by a user of this library, or at least the order of nodes within any given category.

My proposal is thus the following:

* Keep the models of a given category in the order that they were registered. This is backwards compatible, assuming the user doesn't rely on the iteration order of `unordered_map`
* Keep the categories in the order that they are inserted in via `registerModel`. Basically, use a `std::vector` instead of a `std::set`. It might be a good idea to add a `sortCategories()` function. This is backwards incompatible, but I think it's valuable. If we want to stay backwards compatible, there should at least be a way to specify the order of the categories.

PR #163, opened 2018-04-30
Keep registration order as order for context menu

Quincunx271:
Implements #147 

I kept `_categories` and `_registeredModelsCategory` as-is. Someone could've called those functions and stored the variables, and removing them, but especially replacing it with a new data-type, would break their code (replacing could silently break in some cases). My ideal would've been to just change the `categories()` function to return the `std::vector` and to change the `registeredModelsCategoryAssociation` to return a separate map type that preserves insertion order (similar to LinkedHashMap in Java; I would've just combined `std::vector` and `std::unordered_map`).

If removing / changing `_categories` or `_registeredModelsCategory` is desired, I'll make the change and update this PR.


include/nodes/internal/DataModelRegistry.hpp
```diff
@@ -57,11 +60,16 @@ class NODE_EDITOR_PUBLIC DataModelRegistry
 
     QString const name = uniqueModel->name();
 
-    if (_registeredModels.count(name) == 0)
+    if (!_registeredModels.count(name))
```
Quincunx271: You might find this less readable, but I've seen people treating the `count` method on `std::map` or `std::set` as "contains", which is why I wrote is as `!_registeredModels.count(name)` here.

Quincunx271:
examples/calculator

This is the registration code (edited to make this comment smaller):

```c++
  auto ret = std::make_shared<DataModelRegistry>();

  ret->registerModel<NumberSourceDataModel>("Sources");
  ret->registerModel<NumberDisplayDataModel>("Displays");
  ret->registerModel<AdditionModel>("Operators");
  ret->registerModel<SubtractionModel>("Operators");
  ret->registerModel<MultiplicationModel>("Operators");
  ret->registerModel<DivisionModel>("Operators");
  ret->registerModel<ModuloModel>("Operators");
```

Before this PR:

image

After this PR:

image

The categories can still be sorted after this PR by calling `ret->sortCategories()`

charlietsao:
Just found that version 2.1.3 still not support manually context menu settings.

PR #164, opened 2018-05-02
Removed NodeDataModel::clone()

facontidavide:
Hi, 

this is the follow up of #135 .

About this comment from @Quincunx271 :

> I thought about this a bit more, and I think it might be beneficial to use our own Factory type rather than std::function. The strongest benefit that I can see is that it makes it possible in the future for us to add things to this "factory", stuff that are static per NodeDataModel subclass.
> This is similar to the class model in Java, where there's a single Class object that represents the type of all objects of that class. Or in C++, it'd be a lot like static member functions of a type.

I am not sure I understood your suggestion.

For me __DataModelRegistry__ is a Factory. What you probably mean is the [Builder Design Pattern](https://sourcemaking.com/design_patterns/builder/cpp/1), am I right?

From my point of view, a lambda or a function pointer are so easy to write for the user that creating a Factory/Builder just seems an overkill.

About allowing the access to static methods: a strong candidate would be NodeDataModel::name(), since it is supposed to be unique and constant. Is there any other method that you believe should be static?

Remember that a static can not be virtual, nor pure virtual. But there are other way to force the user to implement them ;) 

In the past I managed to solve the "static vs virtual" problem using the Curiously Recursive template Pattern.


facontidavide:
Oops. Apparently visual studio is not happy :(

Does anyone know why std::function won't accept the lambda conversion?

facontidavide:
Fixed...

Quincunx271:
You are right. It's not the Factory Pattern, but seems to be the Builder Pattern (I'd need to brush up on design patterns to really know). Here's a sketch of what I was thinking of:

```c++
class NodeDataModelSpec {
public:
    // Actually create the NodeDataModel
    virtual std::unique_ptr<NodeDataModel> create() const = 0;
    // Other information about the specific kind of NodeDataModel
    virtual QString name() const = 0;
    // ...
};
```

The more I think about this, the more I think that it's not necessary. I don't think we need to do this, because I don't think it makes much of a difference, and because it's possible to add this later in a way that's not API breaking if we do want it.

facontidavide:
By the way; I don't want to make this too complicated, but I have found a way to detect at compile time if a NodeModel about to be registered implemented the static method __QString Name()__.

If it does, the register use that instead of the virtual method name(), that require the creation of an instance of the Node.

Let me know if it is too complicated... old user code should work just fine, we are not breaking the API.

Quincunx271:
It's a nice idea. I think it might be better as it's own PR, though. There's multiple things we could consider for doing this to, and I think it'd be best to do it for all of them if we do it at all.

There's an easier way to write `HasStaticMethodName` with C++11:

```c++
template <typename T, typename = void>
struct HasStaticMethodName
    : std::false_type
{};

template <typename T>
struct HasStaticMethodName<T,
        typename std::enable_if<std::is_same<decltype(T::Name()), QString>::value>::type>
    : std::true_type
{};
```

examples/example2/TextSourceDataModel.hpp
```diff
@@ -35,13 +35,13 @@ class TextSourceDataModel : public NodeDataModel
...
+  static QString
+  Name()
   { return QString("TextSourceDataModel"); }
 
-  std::unique_ptr<NodeDataModel>
-  clone() const override
-  { return std::make_unique<TextSourceDataModel>(); }
+  QString
+  name() const override
+  { return TextSourceDataModel::Name(); }
```
paceholder: Should we be consistent with the style and use the `name` instead of `Name` ?

paceholder:
What is the benefit of allowing both static or non-static `name()` member functions?
Backward compatibility?

I'd stick with just one style. Or maybe I miss something?

Thanks

Dmitry

facontidavide:
yes, backward compatibility only. I guess you have a lot of users ;)
I personally prefer the static because:

- it is a property of the class, not the instance.
- it has a strong guaranty that it can never change.
- you are not obliged to create an instance of the node during the registration.

Up to you if you want to keep only one of the two styles :+1:

examples/example2/TextSourceDataModel.hpp
facontidavide: The name need to be different. We may change __Name()__ to __staticName()__.

Or any other you suggest...

paceholder:
Well, it was just for my understanding. I agree, we can keep both. Maybe deprecate the non-static later.

paceholder:
Thanks for the PR, a great improvement.  I'll merge it. 
Could you please just change the static function name `Name()` -> `name()`.

facontidavide:
No, I can't. the **static** and **non-static-virtual** can NOT have the same name :(

paceholder:
Stupid me. Sorry.

Quincunx271:
I don't think we should deprecate the non-static `name()`; with it, we have the ability to load models at runtime

PR #171, opened 2018-05-11
Improve registry / creator code

Quincunx271:
Simplifies the registration code, removing duplication, by using tag dispatching, and minimizing the compile-time branch code.

The only difference between the two `registerModelImpl` functions was the way the name was being computed, so rather than using `registerModelImpl` to do the compile-time branching, make a new `computeName` function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,8 +33,8 @@ matrix:
       compiler: gcc
       env:
         - CXX=g++-5 CC=gcc-5 QT=5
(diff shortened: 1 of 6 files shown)


Quincunx271:
I also made the `DataModelRegistry` outlive the Connections and Nodes, so that e.g. a NodeDataModel could have a regular pointer (not shared) into something from its creator.

paceholder:
I fixed a couple of valgrind bugs and merged it to master. Thanks.

Issue #174, opened 2018-05-14
Add NodeDataModel "slots" for connections added/removed


Quincunx271:
Currently, if I have a `MyDataModel : NodeDataModel`, I can be informed when the input data changes with `setInData(...)`, and I can be informed when someone requests my output data with `outData(...)`. However, I've encountered a need to know just when new connections are added. I'm proposing adding:

```c++
virtual
void
inputConnectionAdded(Connection const&);

virtual
void
inputConnectionRemoved(Connection const&);

virtual
void
outputConnectionAdded(Connection const&);

virtual
void
outputConnectionRemoved(Connection const&);
```

The exact signatures / names aren't that important, but it's important that the passed information includes:

* The port it's being connected to.
* The other node and its port that we are being connected to.

paceholder:
As I understand, you want to add the functions to the `NodeDataModel` and trigger them from the `Node`? This makes sense. Could you please go ahead and PR it?

Thanks

Dmitry

Quincunx271:
I found a workaround: `connect(this, &FlowScene::signal, this, [this](Connection const& c) { slot(c); });` instead of `connect(this, &FlowScene::signal, this, &FlowScene::slot);`

paceholder:
Merged as #211

PR #175, opened 2018-05-15
Add Connection Created/Destroyed slots to NodeDataModel 

Quincunx271:
Addresses #174 

This adds the following virtual functions to `NodeDataModel`:

* `inputConnectionCreated(Connection const&)`
* `inputConnectionDeleted(Connection const&)`
* `outputConnectionCreated(Connection const&)`
* `outputConnectionDeleted(Connection const&)`

It *might* be more consistent to just have `connectionCreated(PortType, Connection const&)`, instead of `inputConnectionCreated` and `outputConnectionCreated`,  but I believe having them as separate functions is the better approach

### DIFF
--- a/src/FlowScene.cpp
+++ b/src/FlowScene.cpp
@@ -560,10 +560,12 @@ FlowScene::
 sendConnectionCreatedToNodes(Connection const& c)
 {
   Node* from = c.getNode(PortType::Out);
-  Node* to = c.getNode(PortType::In);
+  Node* to   = c.getNode(PortType::In);
 
-  from->nodeDataModel()->outputConnectionCreated(c);
-  to->nodeDataModel()->inputConnectionCreated(c);
+  if (from != nullptr && to != nullptr) {
+    from->nodeDataModel()->outputConnectionCreated(c);
+    to->nodeDataModel()->inputConnectionCreated(c);
+  }
 }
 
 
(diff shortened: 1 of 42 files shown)


Quincunx271:
Don't merge this yet; I may have got the connections backwards

Quincunx271:
AppVeyor is failing. Qt 5.7 no longer exists, and it looks like it would be a good idea to specify a *build worker image*.

https://www.appveyor.com/docs/build-environment/#qt

src/FlowScene.cpp
```diff
@@ -45,6 +47,15 @@ FlowScene(std::shared_ptr<DataModelRegistry> registry,
...
   setItemIndexMethod(QGraphicsScene::NoIndex);
+
+  // Note: the weirdness of using a lambda instead of a member function pointer is
+  // due to a bug in Qt 5.2.
+
+  // This connection should come first
+  connect(this, &FlowScene::connectionCreated, this, [this](Connection const& c) { setupConnectionSignals(c); });
+
+  connect(this, &FlowScene::connectionCreated, this, [this](Connection const& c) { sendConnectionCreatedToNodes(c); });
+  connect(this, &FlowScene::connectionDeleted, this, [this](Connection const& c) { sendConnectionDeletedToNodes(c); });
```
Quincunx271: Using a lambda instead of

```c++
connect(this, &FlowScene::connectionDeleted, this, &FlowScene::sendConnectionDeletedToNodes);
```

The latter works just fine, except the minimal version of Qt we support is 5.2, and there's a bug where it rejects this code (it has a failing static-assertion which is a false positive)

paceholder:
Cleaned-up, rebased, squashed and merged as #211
Thank you!

PR #176, opened 2018-05-17
Fix Appveyor build

Quincunx271:
The appveyor build was failing due to Qt 5.7 not being on the image we were using (https://www.appveyor.com/docs/build-environment/#qt). I specified a *build worker image* and changed the Qt version from 5.7 to 5.6.3.


paceholder:
The image was specified directly on the appveyor website. But it is better to have it in the config file.

image

Issue #165, opened 2018-05-02
Automated Testing


Quincunx271:
This library currently doesn't have any automated tests (unit tests, integration tests, etc). It has the examples, which function a bit like integration tests in that we can check whether the changes to this library breaks the build for the examples. However, there are no tests that the behavior actually is logical.

The first thing we'd have to decide on is what unit testing library/framework we want to use. QtTest is a thing, but I personally found it somewhat tedious to use. My personal favorite C++ unit testing library is [Catch2](https://github.com/catchorg/Catch2). Note that if you set it up right, you can test the widgets, including signals and the link, using whatever unit testing library (as long as the library allows a custom `main` function).

Adding unit tests would be a rather large endeavor, so I'd recommend adding small pieces at a time. If we decide to do this, it would probably also be a good idea to require future pull requests to have unit tests.

paceholder:
I used Catch2 in [my other projects](https://github.com/paceholder/component-manager/tree/master/tests).

I wasn't sure about GUI testing. But if you say it is possible, I'll give it a try ;)

Quincunx271:
So here's something I've found: in order to test some GUI aspects, you have to make a `QApplication`. Unfortunately, this doesn't work with Travis AFAIK, since Travis doesn't have a GUI to connect to. The alternative I've found is to use `QCoreApplication`, but if you use that, you'll get segfaults when you try to use unsupported operations.

Testing with `QApplication` works just fine on my machine, but not on Travis.

paceholder:
I think we can close it as some tests were implemented. Thanks for your help.

PR #177, opened 2018-05-18
Automated tests

Quincunx271:
Starting #165 

This PR adds automated testing via Catch, plus tests for just a couple of things in order to ensure this works:

* on `DataModelRegistry`, check the registration can get the name of the registered model via `name()` and via static `Name()`
* Check that nodes can be dragged (maybe this test is redundant, but it was important to debug this branch, since it requires a GUI on the CI machine).

The way I named the `TEST_CASE`s may not be the best. If you have suggestions, I can update them.

In order to use Qt things that require a `QApplication`, make such tests under the `[gui]` tag and write `auto app = applicationSetup();` (after `#include "ApplicationSetup.hpp"`), which gives you a `std::unique_ptr<QApplication>`. If the test case needed a `QApplication`, but the writer forgot to write that, there will be segfaults (AFAIK, that's just how Qt works)

There's this `[interface]` tag I put on the `DataModelRegistry` test-case. It was intended to be for things in the interface of the library, but tbh I'm not entirely sure why I thought that warranted a tag...

MinGW is specified to *not* build tests. This is because MinGW doesn't support `std::to_string`, and Catch requires C++11.


Quincunx271:
I'm not sure why this didn't trigger an appveyor build. Is it because the file in this repo is `.appveyor.yml` and not `appveyor.yml`?

.travis.yml
```diff
@@ -43,11 +43,27 @@ before_install:
...
+  - cmake .. -DBUILD_TESTING=OFF -DCMAKE_INSTALL_PREFIX="${DEPS_DIR}/install"
+  - cmake --build . -- install
+  - export CMAKE_PREFIX_PATH="${DEPS_DIR}/install/:$CMAKE_PREFIX_PATH"
+  - popd; popd
 
 script:
   - mkdir build
   - cd build
   - cmake -DCMAKE_VERBOSE_MAKEFILE=$VERBOSE_BUILD .. && make -j
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then xvfb-run --server-args="-screen 0 1024x768x24" ./test/test_nodes; fi
```
paceholder: wow, that's a cool trick

.appveyor.yml
```diff
@@ -19,20 +24,44 @@ configuration:
...
+  - git clone https://github.com/catchorg/Catch2.git
+  - cd Catch2
+  - mkdir build
+  - cd build
+  - cmake "-G%GENERATOR%" -DBUILD_TESTING=OFF ..
+  - cmake --build . --target install
+  - set Catch2_DIR=%DEPENDENCIES=%cd%\install\lib\cmake\Catch2
+  - cd ..
+  - cd ..
+  - cd ..
```
paceholder: It might have been much easier to pull the single-header catch into the sources.

paceholder:
Maybe appveyor was offline... Now it seems to work again.

.appveyor.yml
Quincunx271: True, it probably would have been easier. I did it this way because the canonical way to declare a dependency in modern CMake is with `find_package`.

PR #181, opened 2018-06-06
Prefer for-loop over lambda

facontidavide:
Please feel free to reject this PR.

I just think that plain old  for loop looks better than lambda in these cases. 

But it is personal taste, so it is ok to disagree :)

Cheers

Davide

### DIFF
--- a/src/ConnectionGraphicsObject.cpp
+++ b/src/ConnectionGraphicsObject.cpp
@@ -100,9 +100,7 @@ void
 ConnectionGraphicsObject::
 move()
 {
-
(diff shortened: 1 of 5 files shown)


Quincunx271:
I think these lambdas make more sense to be refactored out into their own methods, rather than wrapping it in a loop that loops over an initializer-list of 2 elements

facontidavide:
@Quincunx271 I guess this is the perfect "[bikeshedding](https://en.wiktionary.org/wiki/bikeshedding)" 

 ;)

PR #185, opened 2018-06-08
Cleaner Cleanup

facontidavide:
Hi,

in my code I frequently create/remove programmatically nodes and connections.

Apparently I am doing something weird (or not) and during the cleanup some "zombie" connection still exist somewhere.

This PR will fix the symptom (crash in **Connection::save()** )  but, for the time being, not the actual illness (connection that still exist even if it was supposed to be deleted). 

In the worst case, this code is equivalent to the old one.

Davide

**P.S.**: at some point we migh start a discussion about making the code more robust using std::weak_ptr instead of raw pointers.

### DIFF
--- a/include/nodes/internal/NodeGeometry.hpp
+++ b/include/nodes/internal/NodeGeometry.hpp
@@ -59,10 +59,10 @@ public:
   setHovered(unsigned int h) { _hovered = h; }
 
   unsigned int
-  nSources() const { return _nSources; }
+  nSources() const;
 
   unsigned int
-  nSinks() const { return _nSinks; }
+  nSinks() const;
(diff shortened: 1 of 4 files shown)


facontidavide:
Couldn't resist, I added a couple more commits ;)

paceholder:
Hi Davide,

I know it took me ages so get to this PR. Is it still actual?

Thanks

Dmitry

facontidavide:
No clue... I haven't been following your last changes... :(

paceholder:
Checked -- the PR is still valid. Thank you

facontidavide:
awesome

PR #193, opened 2018-10-09
Automatically get Catch2 if needed

Quincunx271:
* Automatically downloads Catch2 if `BUILD_TESTING` is on and Catch2 is not otherwise found.
* Fix usage of Catch2
  * Use `#include <catch2/catch.hpp>` (Catch2 v2.3.0+)
  * Use `Catch2::Catch2` (Catch2 v2.3.0+)
* Fix appveyor build
  * MinGW can now build tests (passed [`-DCATCH_CONFIG_NO_CPP11_TO_STRING`](https://github.com/catchorg/Catch2/blob/master/docs/configuration.md#c11-toggles))
    * Running the tests fail, though...
* Updated project options
  * Play nicely with `add_subdirectory(nodeeditor)` workflow.
  * Add a `NE_FORCE_TEST_COLOR` option which, when turned on, forces the Catch2 executable to output with color. Useful for running ctest from the commandline
  * If we are the root project, set the `bin` and `lib` directories. We don't want to interfere with anyone else's `CMAKE_*_OUTPUT_DIRECTORY` settings, but it's okay if we do so for ourselves.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -5,15 +5,13 @@ environment:
     - GENERATOR : "Visual Studio 14 2015 Win64"
       QTDIR: C:\Qt\5.6.3\msvc2015_64
       PLATFORM: x64
-      BUILD_TESTING: ON
     - GENERATOR : "Visual Studio 14 2015"
       QTDIR: C:\Qt\5.6.3\msvc2015
       PLATFORM: Win32
-      BUILD_TESTING: ON
     - GENERATOR : "MinGW Makefiles"
       QTDIR: C:\Qt\5.6\mingw49_32
       PLATFORM: x86
-      BUILD_TESTING: OFF
+      CMAKE_CXX_FLAGS_INIT: -DCATCH_CONFIG_NO_CPP11_TO_STRING
 
 image: Visual Studio 2015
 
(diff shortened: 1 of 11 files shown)


.appveyor.yml
```diff
@@ -5,15 +5,13 @@ environment:
...
-      BUILD_TESTING: ON
     - GENERATOR : "Visual Studio 14 2015"
       QTDIR: C:\Qt\5.6.3\msvc2015
       PLATFORM: Win32
-      BUILD_TESTING: ON
     - GENERATOR : "MinGW Makefiles"
       QTDIR: C:\Qt\5.6\mingw49_32
       PLATFORM: x86
-      BUILD_TESTING: OFF
+      CMAKE_CXX_FLAGS_INIT: -DCATCH_CONFIG_NO_CPP11_TO_STRING
```
paceholder: Is this thing documented anywhere?

paceholder:
Thanks a lot for the PR

.appveyor.yml
Quincunx271: https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS_INIT.html

VVD:
Download during build is bad practice.

PR #194, opened 2018-10-15
Fix dereferencing nullptr

sindney:
Dereferencing nullptr node causes crash on osx’s clang.


paceholder:
Good point. Thank you.

Issue #160, opened 2018-04-27
Consider using Q_SIGNALS, Q_SLOTS, Q_EMIT, etc


Quincunx271:
Some C++ libraries out there use the words `signals`, `slots`, etc. as identifiers (e.g. some of Boost). In Qt, these are considered keywords, which makes Qt incompatible with those libraries. However, [Qt lets the user set an option](http://doc.qt.io/qt-5/signalsandslots.html#using-qt-with-3rd-party-signals-and-slots) to decide to turn off the `signals`, `slots`, etc. keywords, and use `Q_SIGNALS`, etc. instead (those macros are always available).

My suggestion is to replace all uses of the Qt keywords with the `Q_KEYWORD` variant, so that users of this library can choose whether they want to set that configuration option.

paceholder:
Merged

PR #199, opened 2018-10-25
Qt no keywords

fredakilla:
see: #160
using Q_SIGNALS, Q_SLOTS, Q_EMIT instead of qt keywords

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -102,6 +102,7 @@ target_compile_definitions(nodes
   PRIVATE
     NODE_EDITOR_EXPORTS
     #NODE_DEBUG_DRAWING
+    QT_NO_KEYWORDS
 )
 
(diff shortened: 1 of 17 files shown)


Quincunx271:
Looks great. The one suggestion I have is to add

```cmake
add_definitions(-DQT_NO_KEYWORDS)
```

to the root CMakeLists.txt, so that building the library will fail if a Qt keyword is used. It may be preferable to write:

```cmake
target_compile_definitions(nodes
  PRIVATE
    QT_NO_KEYWORDS
)
```

Which would accomplish the same thing, but only for the `nodes` target instead of all targets defined under this project.

fredakilla:
right, thanks.

paceholder:
Looks like a reasonable improvement. Thank you

PR #204, opened 2018-12-18
Fix coredump, memory leaks and uses after free

andrejlevkovitch:
Here was big problem with uncorrect uses boundingRect - it changes without call prepareGeometryChange. So I have core dump and some other errors (and memory leaks), which I fixed.


paceholder:
Hi,

Did you change your mind with your PR? Or it still needs some work to be done?

Thanks

Dmitry

Issue #207, opened 2019-01-23
SHARED flag in add_library prevents from building as STATIC


jbltx:
In the CMakeLists.txt file at line 75: 
https://github.com/paceholder/nodeeditor/blob/d4b83952c62b41ef1ad61dbec27639f409b9a6b8/CMakeLists.txt#L75

`SHARED` flag is set and prevents the use of CMake's `BUILD_SHARED_LIBS` to build as `STATIC` or `SHARED`.

From CMake documentation: 

> Global flag to cause add_library to create shared libraries if on.

> If present and true, this will cause all libraries to be built shared unless the library was explicitly added as a static library. This variable is often added to projects as an OPTION so that each user of a project can decide if they want to build the project using shared or static libraries.

They talk about STATIC flag but not SHARED ones, unfortunately if SHARED is set the same prevention is done.

The best way to resolve it is to set the CMake flag as an option (and set to ON by default), as mentioned by CMake doc.

Thank you.

PR #208, opened 2019-01-23
Add cmake option to build as static or shared lib

jbltx:
Fix the issue https://github.com/paceholder/nodeeditor/issues/207


paceholder:
Thank you -- merged

PR #210, opened 2019-02-03
Added the new signal FlowScene::nodePlaced().

signmotion:
Reason: For signal `FlowScene::nodeCreated()` a node has zero position.


paceholder:
Rebased & merged manually

PR #213, opened 2019-03-02
Many in connections

ExMix:


Daguerreo:
I merged it in my fork, seems to work like a charm. As for now the approach seems more reliable of adding dynamic ports.

Daguerreo:
I noticed this PR broke the converters. ModuloModel in calculator example stop working after this.

Daguerreo:
I figured out. It was the converter in Connection passed by copy which cause the shared_ptr to be deleted after been sent to NodeDataModel::setInData().

PR #216, opened 2019-03-11
Add embed/dembed widget posibility

samiavasil:
Add possibility to embed and deembed (detach) Node widget from scene. From popup menu on right button click you can select embedding state. When the widget is detached you can change it size and if you embed again the Node resized to new size.


nolankramer:
Please see https://github.com/paceholder/nodeeditor/pull/374 @samiavasil. It might be prudent to build your changes on top of that PR once it's merged.

PR #223, opened 2019-04-06
 Connection geometry: adequate rounding around the OY axis

andrej22116:
Hi! I use Your solution in my project and I am bored with bends similar to each other in the OY axis when the input port is further than the output port. **I fixed it!** 💃 

Before:
Annotation

After:
Annotation (1)


paceholder:
Hi Andrei,


thank for your PR, indeed, the connections look better now.

Thanks

Dmitry

PR #227, opened 2019-04-18
Fix #198 Avoid connecting nodes causing a loop

### LINKED ISSUE #198: BUG(regression): feedback loops cause crash
#111 expose and fix crash when you connect node output to self input (loop).

However it's still possible to make a loop through graph node and it will crash, like this :
nodecrash

Any idea to detect and forbid cyclic connection ?

Daguerreo:
Fix #198 

Creating a loop between ports cause crash. This is the implementation suggested by @sindney


Daguerreo:
@paceholder I've just noticed my whole forks is being mapped by youtube in this PR I don't know why.

PR #236, opened 2019-05-17
Make NodeDataType to be extendable by open/closed principle

nuustik:
This is needed when DataType is not a simple data object,
but a larger and smarter object that should not be copied by value.
Having it as a pointer with virtual operators allows the model to
extend it and control the operations done with it (NodeConnectionInteraction::canConnect, etc),
where TypeConverter is not possible to use.

### DIFF
--- a/examples/calculator/DecimalData.hpp
+++ b/examples/calculator/DecimalData.hpp
@@ -19,10 +19,10 @@ public:
     : _number(number)
   {}
 
-  NodeDataType type() const override
+  std::shared_ptr<NodeDataType> type() const override
   {
-    return NodeDataType {"decimal",
-                         "Decimal"};
+    return std::make_shared<NodeDataType>("decimal",
+                         "Decimal");
   }
 
   double number() const
--- a/examples/calculator/IntegerData.hpp
+++ b/examples/calculator/IntegerData.hpp
@@ -19,10 +19,10 @@ public:
     : _number(number)
   {}
 
-  NodeDataType type() const override
+  std::shared_ptr<NodeDataType> type() const override
   {
-    return NodeDataType {"integer",
-                         "Integer"};
+    return std::make_shared<NodeDataType>("integer",
+                         "Integer");
   }
 
   int number() const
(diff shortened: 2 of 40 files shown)


Daguerreo:
Shouldn't DataType represent just the "type" of the data and not data itself? A "larger and smarter object" isn't just a NodeData?

nuustik:
Our use case is that one data type needs to accept different other data types and we want to control this from the model. We cannot use TypeConverters and we have thousands of types and creating converters from all the combinations would be horrendous task and would eat a lot of memory in the register. Besides that we actually do not need to convert any data, we just need to control if the connection can be made or not.
If NodeDataType would be a pointer we could then inherit and override the equals operator to not just compare the ids but to compare lists, or id existence in the acceptance list, e.g. It would allow having more control over connection creating.

Daguerreo:
> If NodeDataType would be a pointer we could then inherit and override the equals operator to not just compare the ids but to compare lists, or id existence in the acceptance list, e.g. It would allow having more control over connection creating.

you can do it the exact same thing with the current NodeDataType implementation.

nuustik:
What do you mean? The NodeConnectionInteraction::canConnect compares IDs of NodeDataType. This is not enough for my use case. I need to override this comparison in my code to make the connections possible. Even if I change NodeEditor to compare structs instead of IDs and add struct comparison functions to NodeEditor I still can't override them in my own code to do smarter comparisons.

nuustik:
If NodeConnectionInteraction::canConnect would use NodeData::sameType method instead of comparing the IDs directly, then I can work with it but I guess this is not wanted.

paceholder:
I think it's possible to assign "custom" `NodeConnectionInteraction` to `FlowScene`.

Then we could ask `NodeConnectionInteraction const * FlowScene::hasCustomNodeConnectionInteraction()`.
If the returned pointer is not `nullptr` we'd use the overridden  function `canConnect`.

Inside you could use some matching table tailored to your application.
```
QMap<id, id> compatibleTypes.
```

PR #244, opened 2019-06-28
Make deleteConnection take const.

Malcohol:
A few months ago a change was made which meant the FlowScene::connectionCreated signal provided a const connection. Unfortunately, this meant I can't later use it to call deleteConnection, which currently expects non-const. This change makes delete connection take a const& connection.


emmenlau:
Seems valid and reasonable. Could this be considered?

PR #252, opened 2019-09-30
Add copy paste

eliemichel:
Copy/Paste nodes with Ctrl+C/Ctrl+V
Copy both as a custom MIME data and as text, so that one can copy from/to a text file.


paceholder:
Implemented in mastert

PR #253, opened 2019-11-11
Clearing scene only if file open doesn't fail

tatatupi:
Currently, if a node topology file fails during load, the scene is cleared anyhow.

This fix only calls `clearScene()` after the file open check is performed.


paceholder:
Good fix, thanks!

PR #254, opened 2019-11-11
Compilation issues fix

tatatupi:
We were having two issues compiling the repository.

1) The `find_package pkg` macro, which was conflicting with vcpkg in Windows. Removing it didn't broke the linux build.

2) Qt version lack of compatibility. The `QFontMetrics` class from 5.13  doens't have `width()` method. It was changed to `horizontalAdvance()`. And I guess they it will keep that way for the next LTS version.


paceholder:
I bumped the supported Qt 5 version to 5.15 and Qt 6 should work now as well.

emmenlau:
> I bumped the supported Qt 5 version to 5.15 and Qt 6 should work now as well.

*yay* Thanks!

PR #255, opened 2019-11-11
Adding nodeclick signal

tatatupi:
For my application, a single click on the node should emit a signal so that I could handle some interface functionalities.


emmenlau:
Looks good and useful to me. Could this be considered?

PR #268, opened 2020-08-12
Dynamic port nodes with saving and loading

willys0:
Small fix to the dynamic port nodes PR https://github.com/paceholder/nodeeditor/pull/209 by @signmotion to work with loading and saving of scenes.

Input/Output connection vectors in NodeState would go out of range when loading scenes containing nodes with dynamically set number of ports. I added additional key/value pairs to the saved JSON files which will determine the number of connections of the restored nodes.

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -1,5 +1,7 @@
 add_subdirectory(connection_colors)
 
+add_subdirectory(dynamic_node)
+
 add_subdirectory(example2)
 
 add_subdirectory(calculator)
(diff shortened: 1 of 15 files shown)


skarnproject:
For some reason with this PR the node fails to update connection end positions if the width of the node changes when port is added.

paceholder:
I think this PR is no longer needed as of version 3. 

The possible workflow would be to save the model's state to json and then all the connection ids.
Upon restoring we would first load the models back, it would restore the proper count of `In` and `Out` ports and then we load the serialized `ConnectionIds`.

See the example `examples/dynamic_ports`. Saving and restoring is implemented there.

PR #274, opened 2020-11-14
FlowView can now be created with an externally defined style

matthew-reid:


include/nodes/internal/FlowView.hpp
```diff
@@ -27,6 +28,8 @@ class NODE_EDITOR_PUBLIC FlowView
 
   void setScene(FlowScene *scene);
 
+  void setFlowViewStyle(const FlowViewStyle* style);
```
paceholder, line 31: Who is the owner of this pointer and who must remove it?
I'd simply pass a plain const reference. It's cheap to copy three colors.

paceholder (commented):
I would better assign a new custom style to the object holding all the styles using `StyleCollection::setFlowViewStyle(..)`

include/nodes/internal/FlowView.hpp
matthew-reid: I changed my approach to expose StyleCollection::setFlowViewStyle in the public interface. Please re-review.

PR #275, opened 2020-11-14
Models can now be unregisted from DataModelRegistry

matthew-reid:
I also made some of the interface methods virtual because I need to override them in my application


paceholder:
Interesting idea but I did not get answers to my question in the pull request

PR #281, opened 2021-01-14
Clazy cleanup

Daguerreo:
Clazy analysis and code cleanup. Except for narrowing and vice-versa, most of warning was about:

warning: overrides a destructor but is not marked 'override' [inconsistent-missing-destructor-override]
warning: signal shouldn't be const [clazy-const-signal-or-slot]
warning: signal arguments need to be fully-qualified  [clazy-fully-qualified-moc-types]
warning: slot arguments need to be fully-qualified [clazy-fully-qualified-moc-types]
warning: enumeration value 'None' not explicitly handled in switch [switch-enum]
warning: Missing emit keyword on signal call [clazy-incorrect-emit]
warning: 'break' will never be executed [unreachable-code-break]
warning: Don't call QList::operator[]() on temporary [clazy-detaching-temporary]
warning: 'qsrand' is deprecated: use QRandomGenerator instead [deprecated-declarations]

### DIFF
--- a/include/nodes/internal/ConnectionStyle.hpp
+++ b/include/nodes/internal/ConnectionStyle.hpp
@@ -14,17 +14,17 @@ public:
 
   ConnectionStyle();
 
-  ConnectionStyle(const QString& jsonText);
+  ConnectionStyle(QString const& jsonText);
 
 public:
 
-  static void setConnectionStyle(const QString& jsonText);
+  static void setConnectionStyle(QString const& jsonText);
 
 private:
 
-  void loadJsonText(const QString& jsonText) override;
+  void loadJsonText(QString const& jsonText) override;
 
-  void loadJsonFile(const QString& fileName) override;
+  void loadJsonFile(QString const& fileName) override;
 
   void loadJsonFromByteArray(QByteArray const &byteArray) override;
 
@@ -32,7 +32,7 @@ public:
 
   QColor constructionColor() const;
   QColor normalColor() const;
-  QColor normalColor(const QString& typeId) const;
+  QColor normalColor(QString const& typeId) const;
   QColor selectedColor() const;
   QColor selectedHaloColor() const;
   QColor hoveredColor() const;
--- a/include/nodes/internal/ConnectionStyle.hpp
+++ b/include/nodes/internal/ConnectionStyle.hpp
@@ -14,17 +14,17 @@ public:
 
   ConnectionStyle();
 
-  ConnectionStyle(QString jsonText);
+  ConnectionStyle(const QString& jsonText);
 
 public:
 
-  static void setConnectionStyle(QString jsonText);
+  static void setConnectionStyle(const QString& jsonText);
 
 private:
 
-  void loadJsonText(QString jsonText) override;
+  void loadJsonText(const QString& jsonText) override;
 
-  void loadJsonFile(QString fileName) override;
+  void loadJsonFile(const QString& fileName) override;
 
   void loadJsonFromByteArray(QByteArray const &byteArray) override;
 
@@ -32,7 +32,7 @@ public:
 
   QColor constructionColor() const;
   QColor normalColor() const;
-  QColor normalColor(QString typeId) const;
+  QColor normalColor(const QString& typeId) const;
   QColor selectedColor() const;
   QColor selectedHaloColor() const;
   QColor hoveredColor() const;
--- a/include/nodes/internal/FlowScene.hpp
+++ b/include/nodes/internal/FlowScene.hpp
@@ -36,7 +36,7 @@ public:
 
   FlowScene(QObject * parent = Q_NULLPTR);
 
-  ~FlowScene();
+  ~FlowScene() override;
 
 public:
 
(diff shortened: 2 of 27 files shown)


paceholder:
A huge work! I'll look though it, thanks!

include/nodes/internal/ConnectionStyle.hpp
```diff
@@ -14,25 +14,25 @@ class NODE_EDITOR_PUBLIC ConnectionStyle : public Style
 
   ConnectionStyle();
 
-  ConnectionStyle(QString jsonText);
+  ConnectionStyle(const QString& jsonText);
```
paceholder, line 17: please put the `const` after the type everywhere. That's how the code was formatted in this project

include/nodes/internal/ConnectionStyle.hpp
```diff
@@ -14,25 +14,25 @@ class NODE_EDITOR_PUBLIC ConnectionStyle : public Style
...
 
 public:
 
-  static void setConnectionStyle(QString jsonText);
+  static void setConnectionStyle(const QString& jsonText);
 
 private:
 
-  void loadJsonText(QString jsonText) override;
+  void loadJsonText(const QString& jsonText) override;
```
paceholder, line 25: This wasn't really necessary due to https://doc.qt.io/qt-5/implicit-sharing.html

include/nodes/internal/FlowScene.hpp
```diff
@@ -106,33 +106,46 @@ class NODE_EDITOR_PUBLIC FlowScene
...
+  nodeCreated(QtNodes::Node &n);
 
   /**
    * @brief Node has been added to the scene.
    * @details Connect to this signal if need a correct position of node.
    * @see nodeCreated()
    */
-  void nodePlaced(Node &n);
+  void
+  nodePlaced(QtNodes::Node &n);
```
paceholder, line 118: The class is defined within the `QtNodes` namespace. What is the point of adding the namespaces everywhere?

include/nodes/internal/FlowScene.hpp
```diff
@@ -106,33 +106,46 @@ class NODE_EDITOR_PUBLIC FlowScene
...
-  void nodeHoverLeft(Node& n);
+  void
+  connectionHoverLeft(QtNodes::Connection& c);
 
-  void nodeContextMenu(Node& n, const QPointF& pos);
+  void
+  nodeHoverLeft(QtNodes::Node& n);
+
+  void
+  nodeContextMenu(QtNodes::Node& n, const QPointF& pos);
```
paceholder, line 148: We must be consistent: do move the function name to the new line or keep it here (as below)?

src/ConnectionGraphicsObject.cpp
```diff
@@ -205,7 +205,7 @@ mouseReleaseEvent(QGraphicsSceneMouseEvent* event)
   event->accept();
```
paceholder, line 206: We can get then view by using `auto view = static_cast<QGraphicsView *>(event->widget());`

paceholder (changes requested):

include/nodes/internal/FlowScene.hpp
Daguerreo: Here I just followed the Clazy warning about that

include/nodes/internal/FlowScene.hpp
Daguerreo: Just you choose. I guessed that was your coding style since I never seen it elsewhere. Anyway the whole project has mixed style so I don't know which one to follow. If I could choose I'd prefer not to spread function signature in muiltiple line for both headers and cpp, except for long parameters, but in the end, your code your rules :)

src/ConnectionGraphicsObject.cpp
Daguerreo: Sorry i didn't get it

include/nodes/internal/ConnectionStyle.hpp
Daguerreo: function-args-by-ref

Warns when you should be passing by const-ref. Types with sizeof > 16 bytes [1] or types which are not trivially-copyable [2] or not trivially-destructible [3] should be passed by ref. A rule of thumb is that if passing by value would trigger copy-ctor and/or dtor then pass by ref instead.

include/nodes/internal/FlowScene.hpp
Daguerreo: fully-qualified-moc-types
Warns when a signal, slot or invokable declaration is not using fully-qualified type names, which will break old-style connects and interaction with QML. 

So not a problem with the codebase itself, but since the back-end could communicate with QML could be a point

Daguerreo (commented):
Replies after single comments

include/nodes/internal/FlowScene.hpp
Daguerreo: The following files use a verbose style:

Connection.hpp, NodeDataModel.hpp, ConnectionGeometry.hpp, ConnectionGraphicsObject.hpp, Node.hpp, NodeGeometry.hpp, NodeGraphicsObject.hpp, NodeState.hpp, Style.hpp, StyleCollection.hpp

The following files use a standardy style:

ConnectionStyle.hpp DataModelRegistry.hpp FlowView.hpp FlowViewStyle.hpp, NodeData.hpp, NodeStyle.hpp, PortType.hpp

These use mixed style:
ConnectionState.hpp, FlowScene.hpp

Since there's no a uniform indentation style, I would leave this one as is, and then I would create another pull request with an unique indentation style (you will tell which one). A style cleanup would greatly benefit the project, especially for samples (I already done it there in my fork).

include/nodes/internal/FlowScene.hpp
paceholder: I see, make sense, thanks for explanation.

include/nodes/internal/FlowScene.hpp
paceholder: good!

Daguerreo:
When you're going to merge it, I will start the other PR with code style (I still don't know which one though).

paceholder:
There is requestet changes, that's why this one hasn't been merged.

paceholder:
Please, re-submit after rebasing and implementing the request.

PR #292, opened 2021-11-07
Take recent catch2 to fix build issue

Malcohol:
The referenced versions of Catch2 were not building on Ubuntu 21.10 due to this issue: https://github.com/catchorg/Catch2/issues/2178

I've updated to the latest release.


paceholder:
Thanks!

PR #298, opened 2022-01-29
Minor improvements in CMake and build

emmenlau:
These changes add very minor improvements in CMake, and fix a warning about a shadowed variable. This should be pretty harmless to merge.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,8 +24,10 @@ option(NE_FORCE_TEST_COLOR "Force colorized unit test output" OFF)
 enable_testing()
 
 if(NE_DEVELOPER_DEFAULTS)
-  set(CMAKE_CXX_STANDARD 14)
-  set(CMAKE_CXX_EXTENSIONS OFF)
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -114,12 +116,18 @@ target_compile_options(nodes
...
+  target_compile_options(nodes
+    PRIVATE
+      $<$<CXX_COMPILER_ID:Clang>:-Wall -Wextra>
+  )
+endif()
 
 target_compile_features(nodes
   PUBLIC
-    cxx_generic_lambdas # Require C++14
+    cxx_generic_lambdas # Require C++14 or newer
```
paceholder, line 130: In the resent versions CMake has added the possibility to write the needed standard directly.
I tend to use this expression:

```
target_compile_features(mylib PUBLIC cxx_std_14)
```

paceholder:
I've merged your changes with minor modifications. You could check it everything work for you as planned. Thanks!

PR #300, opened 2022-02-15
Datatype comparison operators

Daguerreo:
Added comparison operator for DataType


Daguerreo:
Sorry, messed up with the change. I'm going to reopen with the correct changes

Issue #138, opened 2018-01-07
Multiple connections


Anton-AuReN:
Is it possible to make multiple connections to the input node?

russelltg:
It doesn't really make sense to have multiple input nodes for a data flow graph, so is not supported. 

However, #117 generalizes this library to be a generic node rendering library, with a default implementation for data flow. I'd suggest you check that out.

paceholder:
It is possible currently in `v3`

PR #302, opened 2022-03-06
Support for multiple incoming connections at a single input port

vhutter:


vhutter:
I modified `example2` to demonstrate the change. Unfortunately, it's a breaking change, since the signature of `NodeDataModel::setInData` was extended by one parameter. However, porting previously written code using the library should be as trivial as mechanically refactoring member functions overriding this, extending it with a similarly typed, unnamed parameter.

paceholder:
Hi @vhutter,

thanks for your pull request.

The featuer you were trying to implement is definitely makes sense and could be useful in some scenarios.
However it cannot be aligned well with the version 2.x desing of the library.
Currently, the graph we are building is supposed to be data-propagating only. And multiple input connections for each port break this semantics.

If we consider general visualation/manipulation tasks for graphs, then any number of connections could be appropriate.

Besides, as you have written, it's a breaking change and I do not want to force hunders of potential users to refactor their code.

Now, regarding the workflow you were aiming at:
In the version 3.x (branch `v3alpha`) I have implemented tools for plain graph visualization, without any nudge to the users to propagate any data. You'd need to derive from `AbstractGraphModel` and implement a policy for ports using the function `AbstractGraphModel::portData(nodeId, portType, portIndex, PortRole::ConnectionPolicyRole)`.
Still, the old approach is also supported. There is a model specialization called `DataFlowGraphModel` that does pretty much the same as the version 2.x

Let me know what you think.

Thanks,

Dmitry

vhutter:
Hi Dmitry,

I've been browsing the issues / PRs history on this repo, and found similar attempts to mine, although I'm not sure they were trying to achieve the exact same thing. I also noticed that it was mentioned already that this doesn't conform to the philosophy of the library (for example in #138), which you said too, but I seem to fail in understanding why. I'm a little afraid to ask at this point, but I tried to search up "data-propagating" or "flow" graphs, and didn't really find a definition matching the structure of the graph in this library. Can you provide some kind of explanation regarding the semantics you mentioned?

> What are you trying to achieve?

I'm working on a node-based editor in the tooling of a game. The editor's purpose would be to visualize multi-layer (tree) data structures of potentially different types, of which height is constant (or at least has an upper bound), but its width is dynamic. What this practically means is that I have a struct in which data members are either instances of a (potentially different) struct, or a container of these. My proposed change is orderedness-agnostic (although from the user's perspective it makes more sense to think of it as an unordered set), unlike implementations changing the number of the ports dynamically, like #209. In my opinion, it would make sense in situations where the user doesn't care about the order of the inputs, only about their "folded" value by some function, or just the fact that they are grouped into one container / category.

> Besides, as you have written, it's a breaking change and I do not want to force hunders of potential users to refactor their code.

I realized it doesn't have to be breaking, I changed it a little, wrapping the original function (which remained pure virtual) into a function depending on the respective connection ID (which is a simple virtual one). This way, the original usage is the same, while users who care about multiple inputs for a slot should override both functions.

> 3.x (branch v3alpha)

Thanks for mentioning it! I checked it out, seems like I would be able to do what I want, but it sadly crashed when I tested it with `calculator`. I also had a compliation error (some missing #includes I think, on msvc). Overall, I think it will be easier for me to just use this version along with my modification, and try merging possible updates if there will be any in the future, on version 2.x. Once version 3 becomes stable, I'll definitely try porting our program to that.

In light of my last commit, I'd suggest you to maybe reconsider merging it, as I don't think it makes the purpose of the library too much different semantically, being a not too much bigger superset of the current version (definitely not as big as version 3). However you decide, thanks for the library, I'll keep using it!

paceholder:
got suddenly sick for a couple of days. 
I think you convinced me. As long as examples work as before, I'll look through the changes very soon and try to merge.
Thanks

paceholder:
I squashed and pushed the branch manually.

PR #311, opened 2022-04-04
Dynamic port count

vhutter:


vhutter:
Hi @paceholder,

I tried to address the issue mentioned under the original PR (#209), taking in account the last fix from #268, and adding a fix to the issue I noted in my commit message.

paceholder:
Implemented in `v3`. There is an example and some documentation on the topic.
Thanks for your contribution

PR #326, opened 2022-10-14
Allow updating node data model after the node being created

bgao:
The size and connections will be recalculated after the node data model update

### DIFF
--- a/include/nodes/internal/Node.hpp
+++ b/include/nodes/internal/Node.hpp
@@ -85,6 +85,8 @@ public:
   NodeDataModel*
   nodeDataModel() const;
 
+  void updateNodeState();
+
 public Q_SLOTS: // data propagation
 
   /// Propagates incoming data to the underlying model.
--- a/src/Node.cpp
+++ b/src/Node.cpp
@@ -176,6 +176,12 @@ nodeState()
   return _nodeState;
 }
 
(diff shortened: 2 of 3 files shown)


paceholder:
Could you please describe your issue in a couple of words?

Thanks

bgao:
> Could you please describe your issue in a couple of words?
> 
> Thanks

Sure.

The old behavior:
Once the Node is created, if the underlying NodeModel is changed, which may change the inputs/outputs of the node, the geometry size and number of connectors of the node in the editor won't change to reflect the NodeModel change.

With the proposed changes:
If  the NodeModel is changed, user can call `updateNodeState()` from Node and the geometry size and corresponding connectors will be updated to reflect the NodeModel change.

paceholder:
Please look at the implementation of `NodeDelegateModel`.
There are several signals:
- `NodeDelegateModel::portsAboutToBeDeleted(...)`
- `NodeDelegateModel::portsDeleted()`
- `NodeDelegateModel::portsAboutToBeInserted(...)`
- `NodeDelegateModel::portsInserted()`

Alternatively, if you are not going to use the data propagation and aim an the simple graph visualization, you could derive your graph model from `AbstractGraphModel` and emit the same signals from this class.

PR #339, opened 2022-12-06
fix: The node resize will now be based on the embedded widget policy

QianMoth:
> You can try using resizeable_images

Win10 
Qt5.15.2&6.3.0
MVSC

**Before:**
You can change the size at will and even make the widget so small that it is invisible.
image

**After:**
Depends on the size property of the inline widget.
You can use `setMinimumSize()` and `setMaximumSize()` to limit the size of the node

If `setFixSize()` is set, even if resizeable is equal to true, it will not be possible to zoom in and out.

*Inadequate:*
When the node is minimized, the mouse continues to move but the node no longer changes. If you move the mouse to zoom in without releasing it, then the starting point of the calculation is the current mouse point.

### DIFF
--- a/src/NodeGraphicsObject.cpp
+++ b/src/NodeGraphicsObject.cpp
@@ -318,12 +318,10 @@ mouseMoveEvent(QGraphicsSceneMouseEvent* event)
 
       oldSize += QSize(diff.x(), diff.y());
 
-      w->setFixedSize(oldSize);
+      w->resize(oldSize);
 
       AbstractNodeGeometry & geometry = nodeScene()->nodeGeometry();
 
-      _proxyWidget->setMinimumSize(oldSize);
-      _proxyWidget->setMaximumSize(oldSize);
       _proxyWidget->setPos(geometry.widgetPosition(_nodeId));
 
       // Passes the new size to the model.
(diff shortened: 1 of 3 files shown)


paceholder:
Thanks for the fix, I'll look into it. Just did not get enought time last days.

src/NodeGraphicsObject.cpp
```diff
@@ -318,13 +318,11 @@ mouseMoveEvent(QGraphicsSceneMouseEvent* event)
...
 
-      w->setFixedSize(oldSize);
+      w->resize(oldSize);
 
       AbstractNodeGeometry & geometry = nodeScene()->nodeGeometry();
 
-      _proxyWidget->setMinimumSize(oldSize);
-      _proxyWidget->setMaximumSize(oldSize);
-      _proxyWidget->setPos(geometry.widgetPosition(_nodeId));
+      // _proxyWidget->setPos(geometry.widgetPosition(_nodeId));
```
paceholder, line 325: Why so? The proxy graphical item must be positioned somehow within the parent item.

paceholder: Oh, I see, it is still done in the function `embedQWidget`

paceholder:
I rebased and squashed the commits manually. Closing this one. Thanks!

Issue #338, opened 2022-12-04
Q_NAMESPACE error when trying to build


soccerdude008:
When I try to build this on Windows with msvc I get the following error:

nodeeditor/include/QtNodes/internal/Definitions.hpp(14): Error: Namespace declaration lacks Q_NAMESPACE macro.

Any clue how to fix that error?

paceholder:
Would be not bad to tell what compiler and Qt version you are using

paceholder:
Qt's docs say that  `Q_NAMESPACE_EXPORT` which we have in the file performs the same function as your missing `Q_NAMESPACE`. For the sake of experimenting try to puth `Q_NAMESPACE` right after the namespace definition.

soccerdude008:
I am running on Windows and using Qt 5.9.7 and MSVC 19.34.31935.0 (Visual Studio 17 2022).

I also tried with MinGW CXX compiler as GNU 6.3.0, and got the same error. 

From a quick search for the docs you mention, it looks like Q_NAMESPACE_EXPORT wasn't introduced until Qt 5.14, so that could be my issue. I tried putting Q_NAMESPACE right after the namespace definition, and it threw a bunch more errors, so I'll try installing a newer version of Qt to see if it resolves it!

paceholder:
> Q_NAMESPACE_EXPORT wasn't introduced until Qt 5.14

Oh, I did not notice that. Sorry, I was initially developing just on 6.x and just lately fixed all inconsistencies to compile on the latest 5.15.xx

I think 5.15 is the only and the last supported version from the 5.x release, it makes sense to update your installation.

zhoutotong:
Hi, I got same error, and my env is: Ubuntu 20.04, Qt Ver: 5.12.
and 5.12 is default Qt version in Ubuntu 20.04, would you solved this problem for Qt5.12？

paceholder:
The continuous integratino on https://www.appveyor.com/ builds this project with MSVS and Qt 5.15

soccerdude008:
@paceholder Sorry I haven't had a chance to try downloading Qt 5.15 yet, but totally expect it will work in that case, so feel free to close out the issue. I guess the only thing you might change in the repo is in the README.rst it lists Qt > 5.2 as a dependency, so maybe update that to Qt >= 5.15?

paceholder:
I updated Readme

rayment:
[...]

PR #340, opened 2022-12-07
add 2 variables to cmake file: QT_PATH, QT_VERSION

zhoutotong:
specify qt version and install path to solved #338

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,6 +27,7 @@ option(BUILD_DOCS "Build Documentation" "${QT_NODES_DEVELOPER_DEFAULTS}")
 option(BUILD_SHARED_LIBS "Build as shared library" ON)
 option(BUILD_DEBUG_POSTFIX_D "Append d suffix to debug libraries" OFF)
 option(QT_NODES_FORCE_TEST_COLOR "Force colorized unit test output" OFF)
+option(QT_VERSION "Specify Qt Version" "")
 
 enable_testing()
 
@@ -44,26 +45,34 @@ add_subdirectory(external)
 
 
 # Find the QtWidgets library
+set(QT_COMPONENTS
+  Core
+  Widgets
+  Gui
+  OpenGL
+)
 find_package(Qt6 QUIET
-  COMPONENTS
-   Core
-   Widgets
-   Gui
-   OpenGL
+  COMPONENTS ${QT_COMPONENTS}
 )
 
 if (NOT Qt6_FOUND)
-  find_package(Qt5 QUIET
-    COMPONENTS
-     Core
-     Widgets
-     Gui
-     OpenGL
-  )
+  if(${QT_VERSION})
+    find_package(Qt5 ${QT_VERSION} QUIET
+      COMPONENTS ${QT_COMPONENTS}
+      PATHS ${QT_PATH}
+    )
+  else()
+    find_package(Qt5 QUIET
+      COMPONENTS ${QT_COMPONENTS}
+      PATHS ${QT_PATH}
+    )
+  endif()
 endif()
 
 if (NOT (Qt6_FOUND OR Qt5_FOUND))
-  message(FATAL_ERRROR "Qt libraries were not found.")
+  message(FATAL_ERRROR " Qt libraries were not found.")
+else()
+  message(STATUS "Compile with Qt version: ${Qt5Core_VERSION_STRING}")
 endif()
 
 if (Qt6_FOUND)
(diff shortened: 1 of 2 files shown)


paceholder:
What is the purpose of specifying a version for Qt 5.xx?
Do you intend to have several Qt5 installations on your computer?

zhoutotong:
> What is the purpose of specifying a version for Qt 5.xx? Do you intend to have several Qt5 installations on your computer?

I try to compile v3.x in Ubuntu 20.04 system, and the default version of Qt in Ubuntu 20.04 is 5.12, it failed with https://github.com/paceholder/nodeeditor/issues/338, then I try to install 5.15 to my system, and it install in that path: $HOME/Qt.

In addition, build version 3. x must use Qt version 5.15 or above.

zhoutotong:
If the minimum Qt version of QtNodes is Qt15, you only need to change the version requirements in readme.rst.

But still hope that the QT_PATH variable can be added, otherwise the manually installed Qt in the Ubuntu environment cannot be found by cmake

paceholder:
> But still hope that the QT_PATH variable can be added, otherwise the manually installed Qt in the Ubuntu environment cannot be found by cmake

It is usually done by defining an enviroment variable `Qt5_DIR`.
It could be, for example: `Qt5_DIR=<QTDIR>/lib/cmake/Qt5`

Alternatively you could pass the root directory of your Qt installation when configuring the cmake.
For example

```
cmake .. -DCMAKE_PREFIX_PATH=$HOME/Qt/5.11.2/gcc_64 
```

zhoutotong:
Thank you for your idea, but I still hope you can specify the Qt version in cmake, for example:
```
find_package(Qt5 5.15...)
```

paceholder:
Sorry, I thought I did it already. In fact, I just modified the README.

paceholder:
@zhoutotong 
Before I restrict the version, could you please invest 10 minutes and try the fix described on this page?
https://stackoverflow.com/questions/50064670/qt-library-link-errors-in-windows
It is called *3rd issue fix*. Maybe we could work-around it easily and do not  block the library users from using an older Qt distribution on their systems.

paceholder:
@zhoutotong  any luck trying the fix?

CMakeLists.txt
```diff
@@ -44,26 +45,34 @@ add_subdirectory(external)
...
+    find_package(Qt5 QUIET
+      COMPONENTS ${QT_COMPONENTS}
+      PATHS ${QT_PATH}
+    )
+  endif()
 endif()
 
 if (NOT (Qt6_FOUND OR Qt5_FOUND))
-  message(FATAL_ERRROR "Qt libraries were not found.")
+  message(FATAL_ERRROR " Qt libraries were not found.")
```
arsdever, line 73:
```suggestion
  message(FATAL_ERRROR "Qt libraries were not found.")
```

CMakeLists.txt
```diff
@@ -44,26 +45,34 @@ add_subdirectory(external)
...
+  OpenGL
+)
 find_package(Qt6 QUIET
-  COMPONENTS
-   Core
-   Widgets
-   Gui
-   OpenGL
+  COMPONENTS ${QT_COMPONENTS}
 )
```
arsdever, line 56: Why do you force `Qt6`? The `QT_VERSION` loses its meaning. Suppose I have `Qt6` installed in parallel with some other version, but I want to use the other one exactly. How do I achieve that? In that context, the `QT_VERSION` variable becomes misleading.

After the comment above, I dived deeper into the PR. As it seems to be, the `QT_VERSION` is not actually the `QT_VERSION`, but the `Qt5` subversion just in case `Qt6` was not found. IMHO the variable name isn't good for that application.

CMakeLists.txt
```diff
@@ -44,26 +45,34 @@ add_subdirectory(external)
...
+      PATHS ${QT_PATH}
+    )
+  endif()
 endif()
 
 if (NOT (Qt6_FOUND OR Qt5_FOUND))
-  message(FATAL_ERRROR "Qt libraries were not found.")
+  message(FATAL_ERRROR " Qt libraries were not found.")
+else()
+  message(STATUS "Compile with Qt version: ${Qt5Core_VERSION_STRING}")
```
arsdever, line 75: Is it the same name for `Qt6`?

CMakeLists.txt
paceholder: Please have a look at the latest master.

In order to specify the Qt version we can use cmake tools, I reperat my old comment here::


It is usually done by defining an enviroment variable `Qt5_DIR`.
It could be, for example: `Qt5_DIR=<QTDIR>/lib/cmake/Qt5`

Alternatively you could pass the root directory of your Qt installation when configuring the cmake.
For example

```
cmake ..

PR #344, opened 2022-12-12
Fix a bug in the scene menu search, update the treeView creation process

QianMoth:
- Before the bug was fixed, entering search content would cause the system to crash because the `treeView` pointer was cleared.

- Use the `QTreeWidgetItem` property `flag()` `Qt::ItemIsSelectable` to determine if it is a parent.

- Use `QTreeWidgetItemIterator` to iterate over children items and hide/show them

### DIFF
--- a/src/DataFlowGraphicsScene.cpp
+++ b/src/DataFlowGraphicsScene.cpp
@@ -76,8 +76,6 @@ createSceneMenu(QPointF const scenePos)
 {
   QMenu* modelMenu = new QMenu();
 
-  auto skipText = QStringLiteral("skip me");
-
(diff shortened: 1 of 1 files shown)


paceholder:
Hello, thanks for your work.
Could you please explain exactly how to reproduce the bug? I did not manage to do it.

QianMoth:
Win10 , in the current master branch, right-clicking brings up a mods menu bar with a search box. When any character is entered it will cause the system to crash.

I don't know if I have expressed myself clearly?

thanks

paceholder:
I was on a wrong branch. Thank, now I see the crash

PR #348, opened 2022-12-16
Optimising the scene menu creation process, remove `CategoriesSet`

QianMoth:
No longer use `registry->categories()` to create parent items

I looked up where `CategoriesSet` is used in v2 and v3 and found that it is only used when creating menus. `RegisteredModelsCategoryMap` is perfectly capable of doing the job of `CategoriesSet`.

So I think the `CategoriesSet` can now be removed.

Please contact me if I have any incomplete considerations. 
Thanks.


paceholder:
Your changes are totally make sense.

What do you think about this one ? #147
And the corresponding PR https://github.com/paceholder/nodeeditor/pull/163

thanks

QianMoth:
- I use `std::vector` to store the `RegisteredModelsCategory`, and since the class seems to be used only to generate the Scene Menu, just traversing it without using lookups, using `std::vector` not only sorts the classes in the order they are registered, but also reduces the performance loss (though not by much)
- I don't think it's a good idea to use `ret->sortCategories()` to sort at the end of the registration model. Because I won't know which model will be loaded last in my [`plugin_extension`](https://github.com/QianMoth/nodeeditor/tree/plugin_extension)branch because of the plugin name change.
- I have used a bool variable to flag whether or not the Menu needs to be sorted when it is generated, but I think the best way to do this would be to place a Button in the Menu to choose whether or not to sort it, and to be able to take effect the next time the Menu is generated.

Also I think it should just sort the subdirectories, for example *Addition* in the first.

QianMoth:
**Sort button function**
image
image

Issue #335, opened 2022-11-28
Plug-in functions


QianMoth:
Hello @paceholder, are you going to implement plugins in version 3? I implemented a simple dynamic link library management based on the second version.
https://github.com/QianMoth/nodeeditor-p.git
https://github.com/QianMoth/PluginDemo.git
I wish there was a better solution, preferably one that doesn't rely on node.dll and uses only a few header files when making plug-ins.

你好，你打算在第三版本实现插件功能么？我基于第二版本实现了一个简单动态链接库的管理。
https://github.com/QianMoth/nodeeditor-p.git
https://github.com/QianMoth/PluginDemo.git
我希望有更好的解决方案，最好在制作插件的时候不需要依赖node.dll，只使用少量的头文件。

isteelcx:
你是不是把翻译软件的中文也粘贴过来了？

isteelcx:
> 

你是不是把翻译软件的中文也粘贴过来了？

isteelcx:
@QianMoth 你是不是把翻译软件的中文也粘贴过来了？

paceholder:
Hi @QianMoth ,

what exactly do you mean under "plugins "? What do your plugins do? 

Thanks for the clarifications

QianMoth:
> Hi @QianMoth ,
> 
> what exactly do you mean under "plugins "? What do your plugins do?
> 
> Thanks for the clarifications

ummm, My idea is that every node, or every type of node (such as Image Load Node), can be made into a plug-in (a '.dll 'file in windows).
I can then use the plug-in manager to invoke the DataModelRegistry function to dynamically add or remove new nodes.
This allows you to add new functionality without changing the code

QianMoth:
Originally I had two plugins, three nodes. When I removed one of the plug-ins called tcp and reloaded it, I was left with only one node called demo. The same goes for adding nodes

https://user-images.githubusercontent.com/55644167/204253465-49c679a2-96c0-4b56-8821-57bbd25b866a.mp4

paceholder:
Thanks for the video. I think your plugin system lies beyond the scope of the current project. Node editor mainly focuses on graphics and does not imply any logic behind your node model. Your work looks like a good extension to the existing code. Can you just import node editor into your project and extend its classes? For example, you could extend the Node Data Model Registry to make it capable of loading dlls 

What do you think?

PR #349, opened 2022-12-16
Plugin extension (#335)

QianMoth:
#335


paceholder:
Thanks, this is an interesting extension.
I guess code still needs some polishing. The example plugin libs do not land to the directory which is gonna be used for loading the plugins. After fixing that I still did not manage to load the plugin, my system was complaining like 'this library is not a valid Qt plugin'. Must be something with the export/import statements. I'll look into it deeper

QianMoth:
Thank you for your affirmation.
I only have windows environment to program now.

I am just a rookie now and have limited experience.
You are welcome to participate in this part of the development and I am sure I can learn more skills.

QianMoth:
> Thanks, this is an interesting extension. I guess code still needs some polishing. The example plugin libs do not land to the directory which is gonna be used for loading the plugins. After fixing that I still did not manage to load the plugin, my system was complaining like 'this library is not a valid Qt plugin'. Must be something with the export/import statements. I'll look into it deeper

I have tested it on Windows 10 with ubuntu 20.04.

nolankramer (approved):

PR #350, opened 2022-12-16
Clang format

### LINKED ISSUE #146: Add a .clang-format file
Clang Format is an auto-formatter for C++ source code. As there are a lot of C++ code formatting styles, it lets you customize your formatting with a *.clang-format* file. I think it would be helpful to add such a file to this repository; that way contributors can autoformat the code they submit to match the style of the repository.

Daguerreo:
.clang-format has been added. 
The reference for the configuration is the default used by Qt here: https://raw.githubusercontent.com/qt-creator/qt-creator/master/.clang-format

The whole codebase has been formatted following these rules. In linux it's very straightforward, just launch

`find . \( \( -name \*.cpp -o -name \*.hpp \) -print0 | xargs -0 -n 1 clang-format-13 -i`

from the repository root directory.

With Qt creator it's possible to auto-indent the file on save with the beautifier plugin. In `Preferences -> Beaufitifer -> General` check the "enable auto format on file save. Then in the tab `clang-format` select `use predefined style: File`. If the local machine does not have clang-format installed, it can be found in the Qt installation in the folder `Tools/QtCreator/libexec/qtcreator/clang/bin/clang-format`. This might be different depending on the Qt Creator version installed. This guide is valid for Linux, I cannot check in other OS, but it should be quite similar.

The CI can be configured to check the proper code format and reject the MR or even auto-correct the code. I don't know how to do that with the github action since I'm using Gitlab. and I think it should be a different task.

Please @paceholder check if you're ok with the current code format.

Closes #146

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,102 @@
+---
+Language:        Cpp
+AccessModifierOffset: -4
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: DontAlign
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterReturnType: None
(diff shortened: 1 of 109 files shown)


paceholder:
@Daguerreo thanks for your work. This PR is on my list, needed to sort out some smaller things.

paceholder:
For my information -- did you introduce some differences compared to the reference file  https://raw.githubusercontent.com/qt-creator/qt-creator/master/.clang-format ?

paceholder:
Re-Formatted manually due to multiple recent changes & merged to master.

Thanks!

PR #354, opened 2022-12-20
Simplifying cmake

QianMoth:
Every time I add or delete a file I forget to change cmake, change it to this and there will be no problem.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,41 +37,31 @@ if(QT_NODES_DEVELOPER_DEFAULTS)
 endif()
 
 if(BUILD_DEBUG_POSTFIX_D)
-  set(CMAKE_DEBUG_POSTFIX d)
+  set(CMAKE_DEBUG_POSTFIX "d")
+  set(CMAKE_RELEASE_POSTFIX "")
+  set(CMAKE_RELWITHDEBINFO_POSTFIX "rd")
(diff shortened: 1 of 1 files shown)


paceholder:
The explicit listing of files has been done on purpose
https://github.com/paceholder/nodeeditor/pull/142

Besides, your proposed changes with the variable `${QT_VERSION_MAJOR}` does not work for some reason. I do not know whether  it depends on cmake, OS or Qt version. In my case (Qt 6.4, MacOS, cmake 3.25) the variable remains empty after finding the Qt libs.

QianMoth:
Thank you for your reply

QianMoth:
But there is macos qt5.15.2 in the github ci and the compilation passes

paceholder:
You are right, let me pull you branch to my computer and see what exactly happens

paceholder:
Opps looks like you removed your branch completely

QianMoth:
> Opps looks like you removed your branch completely

i restore it

QianMoth:
Hi @paceholder , One more thing that needs to be fixed , Now in cmake you use `FATAL_ERRROR` with the extra letter `R`

paceholder:
Sorry for closing this PR immediately, I see that your changes work now. Dunno what was wrong with my attempts before.

Please have a look: https://github.com/paceholder/nodeeditor/pull/355
I took all your proposed changes, move an MSVC flag and kept the whole list of files

PR #355, opened 2022-12-20
Simplify find_package function calls for Qt5 and Qt6

paceholder:

PR #356, opened 2022-12-20
[update simplify-qt-find] minimun version set to 5.11.0

QianMoth:
https://github.com/paceholder/nodeeditor/pull/355#discussion_r1053805441
1. Minimum version set to 5.11.0; 
2. delete core gui; 
3. ignore `build*/` , The Qt Creater will generate the `build-...` Folder

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,7 +47,7 @@ endif()
 add_subdirectory(external)
 
 find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Widgets)
-find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Widgets OpenGL)
+find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Core Widgets Gui OpenGL)
 message(STATUS "QT_VERSION: ${QT_VERSION}, QT_DIR: ${QT_DIR}")
 
 if (${QT_VERSION} VERSION_LESS 5.11.0)
@@ -151,7 +151,9 @@ target_include_directories(QtNodes
 
 target_link_libraries(QtNodes
   PUBLIC
+    Qt${QT_VERSION_MAJOR}::Core
     Qt${QT_VERSION_MAJOR}::Widgets
+    Qt${QT_VERSION_MAJOR}::Gui
     Qt${QT_VERSION_MAJOR}::OpenGL
 )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -47,11 +47,11 @@ endif()
 add_subdirectory(external)
 
 find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Widgets)
-find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Core Widgets Gui OpenGL)
+find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Widgets OpenGL)
 message(STATUS "QT_VERSION: ${QT_VERSION}, QT_DIR: ${QT_DIR}")
 
-if (${QT_VERSION} VERSION_LESS_EQUAL 5.9.0)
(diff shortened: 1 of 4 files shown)


paceholder:
`5.11` seems to be reasonable. 
For examle, the earliest yet supported Ubuntu distribution is 22.10. It uses Qt 5.15 (if I do not mix up anything)

CMakeLists.txt
```diff
@@ -47,11 +47,11 @@ endif()
 add_subdirectory(external)
 
 find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Widgets)
-find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Core Widgets Gui OpenGL)
+find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Widgets OpenGL)
```
paceholder, line 50: Why the modules have been removed?

paceholder: Are Core and Gui modules linked automatically?

QianMoth: It seems to have been included and is able to compile and run properly.

paceholder: Actually they are included everywhere 
image

paceholder: image

QianMoth: https://doc.qt.io/qt-6/cmake-variable-reference.html

According to this document I printed the message of `Qt6Widgets_INCLUDE_DIRS`. It may be the reason why the `gui` and `core` are deleted but the header file contains no errors.

I can change that part back if you wish.

`message(${Qt6Widgets_INCLUDE_DIRS})`

```bash
D:/Software/Qt/6.2.1/msvc2019_64/include/QtWidgetsD:/Software/Qt/6.2.1/msvc2019_64/includeD:/Software/Qt/6.2.1/msvc2019_64/include/QtCoreD:/Software/Qt/6.2.1/msvc2019_64/include/QtGui
```

paceholder: I am pretty much sure that here works a CMake mechanics. It defines targets with public properties. QtWidgets target must depend on QtCore, therefore it "inherits" all the include properties, i.e. all the include search paths ftrom QtCore

paceholder: The same must be happening with linking. 

I dunno what is better: to list all the dependencies explicitly or to use the cmake specifics

QianMoth: So let's keep it the same

include/QtNodes/internal/Definitions.hpp
```diff
@@ -13,7 +13,7 @@
 
 namespace QtNodes
 {
-Q_NAMESPACE_EXPORT(NODE_EDITOR_PUBLIC)
+NODE_EDITOR_PUBLIC Q_NAMESPACE
```
paceholder, line 16: Does it actually work? I have no chance of testing it on an older versions of Qt on Windows

paceholder, line 16: I mean, maybe it compiles, but can we load dlls and run examples?   I need to fix unit-tests asap

QianMoth: Yes, I've run both caculater and simple_node and added node

QianMoth: Sorry, I only have a Win10 environment。Under Win10 there is no problem.

https://github.com/paceholder/nodeeditor/pull/

Here are the results of my tests on various versions of qt under Win10 and running the example

QianMoth: I have now tested the latest simplify-qt-find branch more thoroughly. On windows 10, qt version 5.11.0.
The following sections are included:
- models create and delete
- connections create and delete
- copy paste 
- drag
- resize

And I ran all the accompanying examples and did their basic operations.

paceholder:
I cleaned up the branch and merged it into master. Thank you!

PR #358, opened 2022-12-27
UndoCommands new feature and fix

### LINKED ISSUE #357: When dragging to move models, cannot drag all the selected models.
This phenomenon exists in v3.
v2 is normal.

QianMoth:
* drag all the selected models (Fix #357), and distinguishes between different drags (with different nodes selected).
* add CreateCommand to create node.
* fix DeleteCommand, If nothing is selected for deletion, mark this DeleteCommand as obsolete.
* fix CopyCommand, Just a copy of connections will crash
* fix PasteCommand, If `"model-name"` does not exist in the pasted content or if the value is not in the registry, it will crash

### DIFF
--- a/src/UndoCommands.cpp
+++ b/src/UndoCommands.cpp
@@ -445,10 +445,8 @@ redo()
 
 MoveNodeCommand::
 MoveNodeCommand(BasicGraphicsScene* scene,
-                NodeId const nodeId,
                 QPointF const &diff)
   : _scene(scene)
-  , _nodeId(nodeId)
   , _diff(diff)
 {
 }
@@ -457,13 +455,19 @@ void
 MoveNodeCommand::
 undo()
 {
-  auto oldPos = 
-    _scene->graphModel().nodeData(_nodeId,
-                                  NodeRole::Position).value<QPointF>();
+  for (QGraphicsItem * item : _scene->selectedItems())
+  {
+    if (auto n = qgraphicsitem_cast<NodeGraphicsObject*>(item))
+    {
+      auto oldPos = 
+        _scene->graphModel().nodeData(n->nodeId(),
+                                      NodeRole::Position).value<QPointF>();
 
-  oldPos -= _diff;
+      oldPos -= _diff;
 
-  _scene->graphModel().setNodeData(_nodeId, NodeRole::Position, oldPos);
+      _scene->graphModel().setNodeData(n->nodeId(), NodeRole::Position, oldPos);
+    }
+  }
 }
 
 
@@ -471,13 +475,19 @@ void
 MoveNodeCommand::
 redo()
 {
-  auto oldPos = 
-    _scene->graphModel().nodeData(_nodeId,
-                                  NodeRole::Position).value<QPointF>();
+  for (QGraphicsItem * item : _scene->selectedItems())
+  {
+    if (auto n = qgraphicsitem_cast<NodeGraphicsObject*>(item))
+    {
+      auto oldPos = 
+        _scene->graphModel().nodeData(n->nodeId(),
+                                      NodeRole::Position).value<QPointF>();
 
-  oldPos += _diff;
+      oldPos += _diff;
 
-  _scene->graphModel().setNodeData(_nodeId, NodeRole::Position, oldPos);
+      _scene->graphModel().setNodeData(n->nodeId(), NodeRole::Position, oldPos);
+    }
+  }
 }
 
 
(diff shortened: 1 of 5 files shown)


QianMoth:
I think it is now complete.
- I can drag and drop all the selected nodes.
- When I drag multiple nodes and then drag a single node, this is distinguished as two commands.
- I can create a node and undo the creation, and will record the parameters you want to save before undoing
- This command will be destroyed on failure to create
- When there are no nodes or connections are selected and then the delete button is pressed, this command will be destroyed

https://user-images.githubusercontent.com/55644167/209745374-0e7bfefc-dd3e-4711-a8bd-10c26d118207.mp4

paceholder:
Thanks for your contribution!

The grouped node move was working in `v2` by calling `QGraphicsItem::mouseMoveEvent(...)`, this line has been unfortunately lost. But now with the undo-command approach we must follow your strategy

src/UndoCommands.cpp
```diff
@@ -292,18 +360,35 @@ redo()
...
-  if (_newSceneJson.empty())
-    return;
-
-  _newSceneJson = makeNewNodeIdsInScene(_newSceneJson);
+  // Ignore if pasted in content that does not generate nodes.
+  try
+  {
+    insertSerializedItems(_newSceneJson, _scene);
+  }
+  catch(...)
```
paceholder, line 368: Do you think it might be better to catch a specific exception?

src/UndoCommands.cpp
```diff
@@ -292,18 +360,35 @@ redo()
 {
   _scene->clearSelection();
 
-  _newSceneJson = takeSceneJsonFromClipboard();
-
-  if (_newSceneJson.empty())
-    return;
-
-  _newSceneJson = makeNewNodeIdsInScene(_newSceneJson);
+  // Ignore if pasted in content that does not generate nodes.
```
paceholder, line 363: ~that~

src/UndoCommands.cpp
```diff
@@ -292,18 +360,35 @@ redo()
...
+  try
+  {
+    insertSerializedItems(_newSceneJson, _scene);
+  }
+  catch(...)
+  {
```
paceholder, line 369: Do we delete all the selected nodes if the "ctrl+v" was unsuccessful ?
Sorry, did not get this branch of the code

src/UndoCommands.cpp
QianMoth: I just used the `FlowScene::restoreNode(...)` method in v2, in the `FlowScene.cpp` file. 

With catch I can also catch the error generated by add connections, although it doesn't seem to be there at the moment. This is as advantageous as I can think of at the moment.

Of course I also tried to modify `void loadNode(...)` to `bool`, if it's `true` it means it was loaded successfully. But this can cause quite a lot of changes.

src/UndoCommands.cpp
QianMoth: I think this is necessary.

In the `calculator` example, If you paste the following, using the no-delete method, you will only get the `Result` node. and it cannot be undone.

So I think it's best to delete all created content, or you can also check if all nodes are creatable first, and then go ahead and create them.

```json
{
    "connections": [
        {
            "inPortIndex": 0,
            "intNodeId": 1,
            "outNodeId": 0,
            "outPortIndex": 0
        }
    ],
    "nodes": [
        {
            "id": 1,
            "internal-data": {
                "model-name": "Result"
            },
            "position": {
                "x": 75,
                "y": -16
            }
        },
        {
            "id": 0,
            "internal-data": {
                "model-name": "N--Source",
                "number": "0"
            },
            "position": {
                "x": -258,
                "y": -83
            }
        }
    ]
}

```

paceholder: I see your point, thanks

src/UndoCommands.cpp
```diff
@@ -292,18 +360,35 @@ redo()
...
+      }
+    }
 
-  offsetNodeGroup(_newSceneJson, _mouseScenePos - averagePos);
+    QJsonArray nodesJsonArray;
+    for (QGraphicsItem * item : _scene->selectedItems())
+    {
+      if (auto n = qgraphicsitem_cast<NodeGraphicsObject*>(item))
+      {
+        graphModel.deleteNode(n->nodeId());
```
paceholder, line 386: I think `setSelected` is missing for the created connections in the function `insertSerializedItems`

QianMoth: Yes, I have amended

paceholder:
I reordered commits and cleaned up the change history. Merged manually, thanks for your work.

PR #360, opened 2023-01-09
Fixed broken context search and hide empty categories from filter.

rayment:
Searching through the context menu was broken after 883e8a0 due to a mis-comparison on node data that is never inserted, returning no results. This has been fixed by correctly comparing against `QTreeWidgetItem::text`.

Furthermore, the filtering system has been changed to hide empty categories, showing only those that are relevant (contain matching nodes).

New filter demonstration

Tested on Gentoo Linux, kernel 6.1.0, Qt 5.15.7


rayment:
Tested also on Windows Server 2022 10.0.20348 Build 20348, Qt 5.15.2.

Fixed filtering on Windows

PR #362, opened 2023-01-12
Modify node's port creation method

zmoth:
This PR is just for the record and you can choose which parts you also agree with.

This is a very tedious update and may not have been well thought out in some cases.

Mainly because the original uses too many unnecessary functions in the port creation section, so my idea is to manage the port section directly with a `std::vector`.

This update also implements dynamic ports and writes them as `dynamic_port2` examples.

The indexed examples were all tested successfully on **windows**, including but not limited to copy and paste, save and load, and data passing.

Here are the main changes.

* use `initModelFormId()` to `addNode()` or `loadNode()`
* `using NodeDataType = QString;`
* add `NodePorts`
* use `_inputPorts` and `_outputPorts` to manage the port


2822907581:
有一些潜在的问题与原来的项目的想法相悖。如下（可能不全，目前只能回忆起这些）： 
1.NodeData及其子类的type属性是动态的，并非一种NodeData只能具有一种固定的type属性。您在NodeDelegateModelRegistry类中添加的映射实现的是一一对应，这与该项目的思路违背。 
2.端口具有的dataType属性与其存储数据的type是分离的，并非dataType等同于type。您似乎忽视了这一点。 
3.DataFlowGraphModel::portData获取PortRole::DataType时，返回的应当是端口的dataType，而不是其所存储数据的type。 
﻿ 
There are a number of potential problems that run counter to the idea of the original project. They are as follows (they may be incomplete, that's all I can recall at the moment): 
1. the type attribute of a NodeData and its subclasses is dynamic, not that a NodeData can only have one fixed type attribute. The mapping you added to the NodeDelegateModelRegistry class implements a one-to-one correspondence, which goes against the idea of the project. 
2. the dataType attribute of a port is separate from the type of data it stores, not dataType is equivalent to type. you seem to have missed this point. 
3. when DataFlowGraphModel::portData gets PortRole::DataType, it should return the dataType of the port, not the type of the data it stores. 
﻿

Issue #364, opened 2023-01-16
BasicGraphicsScene::nodeMoved is not being emitted anywhere


rayment:
I require the use of `BasicGraphicsScene::nodeMoved` (via. `DataFlowGraphicsScene`) for my project as I wish to determine when nodes have been moved. When I checked the code, the signal is not being emitted anywhere at all.

Under the assumption that this signal _should_ be implemented and not left to the user, I have thought of two simple solutions, though each are diametric to one another:

- Emit the signal directly in `BasicGraphicssScene::onNodePositionUpdated`. This produces a signal on every draw frame that the node is being dragged on. It's also redundant because `nodePositionUpdated` is already a signal.
- Emit the signal from a new slot called `BasicGraphicsScene::onNodeClicked`. This can produce a signal once, when the node has been dragged after at least one frame, and the mouse has been released.

I liked the second solution better, and have implemented it in my fork [here](https://github.com/rayment/nodeeditor/commit/a02f2268d5d1cdb84a682738d47359b7722adebe). Alternatively, my implementation for the first solution is as follows:

```c++
void BasicGraphicsScene::onNodePositionUpdated(NodeId const nodeId)
{
    auto node = nodeGraphicsObject(nodeId);
    if (node) {
        auto pos = _graphModel.nodeData(nodeId, NodeRole::Position).value<QPointF>();
        node->setPos(pos);
        node->update();
        Q_EMIT nodeMoved(nodeId, pos);
    }
}
```

Please let me know if this change is useful, and if so, I'll open a PR.

rayment:
Looks like this issue is much bigger than this. After the version upgrade to `3.x.x`, a lot of these slots are no longer called including `inputConnectionCreated/Deleted` and `outputConnectionCreated/outputConnectionDeleted` after `src/FlowScene.cpp` was removed (see [PR 211](https://github.com/paceholder/nodeeditor/pull/211) when it was introduced).

paceholder:
Thanks for addressing it. Indeed, some slots might have gone lost during refactoring (unintentionally). 
I'll have a look at your PR

PR #366, opened 2023-01-26
Re-implement node connection creation/deletion slots and node movement signal.

rayment:
The following slots have been fixed as they were never being called:
- `NodeDelegateModel::inputConnectionCreated`
- `NodeDelegateModel::inputConnectionDeleted`
- `NodeDelegateModel::outputConnectionCreated`
- `NodeDelegateModel::outputConnectionDeleted`

The following slot has been implemented:
- `BasicGraphicsScene::onNodeClicked`

The following signal has been fixed as it was never being emitted:
- `BasicGraphicsScene::nodeMoved`

A nuance concerning `nodeMoved`, unlike `AbstractGraphModel::nodePositionUpdated` which is called every frame of movement and captured by `BasicGraphicsScene::onNodePositionUpdated`, `nodeMoved` will only be called when the mouse is released, locking the node into its new position. Similarly, `onNodeClicked` is only called when the mouse is released. See [Issue 364](https://github.com/paceholder/nodeeditor/issues/364) for more information.

Note also that [TestFlowScene.cpp](https://github.com/paceholder/nodeeditor/blob/master/test/src/TestFlowScene.cpp#L32) is old and no longer works in v3, but does contain some example uses of the fixed slots.

### DIFF
--- a/src/DataFlowGraphModel.cpp
+++ b/src/DataFlowGraphModel.cpp
@@ -134,7 +134,7 @@ void DataFlowGraphModel::addConnection(ConnectionId const connectionId)
 {
     _connectivity.insert(connectionId);
 
-    Q_EMIT connectionCreated(connectionId);
+	sendConnectionCreation(connectionId);
 
     QVariant const portDataToPropagate = portData(connectionId.outNodeId,
                                                   PortType::Out,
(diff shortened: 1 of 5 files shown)


src/DataFlowGraphModel.cpp
```diff
@@ -134,7 +134,7 @@ void DataFlowGraphModel::addConnection(ConnectionId const connectionId)
 {
     _connectivity.insert(connectionId);
 
-    Q_EMIT connectionCreated(connectionId);
```
paceholder, line 137: We can't simply stop emitting this signal.
It is attached to the `BasicGraphicsScene` which in turn repaints all the connected nodes.

paceholder:
A general remark: could you please format the code using .clang-format file from the root directory? Or at least fix the indentaion (4 spaces). Thanks

paceholder:
I fixed the formatting manually and merged

rayment:
> I fixed the formatting manually and merged

Sorry, I've been at work and did not have time to reply. I'll make sure to format it next time.

PR #368, opened 2023-02-01
Add fontFamily options

zh18588441437:
Hi 
I'm learning nodeeditor and looking to do something with it, this commit adds a font selection field

### DIFF
--- a/include/QtNodes/internal/NodeStyle.hpp
+++ b/include/QtNodes/internal/NodeStyle.hpp
@@ -1,4 +1,4 @@
-#pragma once
+﻿#pragma once
 
 #include <QtGui/QColor>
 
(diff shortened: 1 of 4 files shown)


include/QtNodes/internal/NodeStyle.hpp
```diff
@@ -27,27 +27,27 @@ class NODE_EDITOR_PUBLIC NodeStyle : public Style
...
-    float Opacity;
+    QColor  NormalBoundaryColor;
+    QColor  SelectedBoundaryColor;
+    QColor  GradientColor0;
+    QColor  GradientColor1;
+    QColor  GradientColor2;
+    QColor  GradientColor3;
+    QColor  ShadowColor;
+    QColor  FontColor;
+    QColor  FontColorFaded;
```
nolankramer, line 38: Please revert changes like these. There is no need to change these lines of code.

As a word of advice - it is usually desirable to only change code that needs to be changed to address the PR. This avoids causing confusion/overload when others review the change.

Additionally, stylistic coding is really up to repo owner as of now. And it seems like a single space between type and variables is the overall style (and it typically is).

include/QtNodes/internal/NodeStyle.hpp
```diff
@@ -27,27 +27,27 @@ class NODE_EDITOR_PUBLIC NodeStyle : public Style
...
+
+    QColor  WarningColor;
+    QColor  ErrorColor;
+
+    float   PenWidth;
+    float   HoveredPenWidth;
+
+    float   ConnectionPointDiameter;
+
+    float   Opacity;
```
nolankramer, line 51: Please revert adding these additional spaces.

resources/DefaultStyle.json
```diff
@@ -14,6 +14,7 @@
     "ShadowColor": [20, 20, 20],
     "FontColor" : "white",
     "FontColorFaded" : "gray",
+    "FontFamily": "Microsoft YaHei",
```
nolankramer, line 17: The default font should be the default font of Qt: "Segoe UI"

(This font ships with Qt, so we know it's safe to use)

nolankramer (changes requested):

PR #370, opened 2023-02-13
Add a signal `modified` for the scene.

zmoth:
This signal is sent when nodes or connections change.

Many editors will have a file modified flag after modifying the file. We can connect this signal to determine which scene has been modified.
image
image

### DIFF
--- a/examples/calculator/main.cpp
+++ b/examples/calculator/main.cpp
@@ -77,8 +77,12 @@ int main(int argc, char *argv[])
 
     auto menuBar = new QMenuBar();
     QMenu *menu = menuBar->addMenu("File");
+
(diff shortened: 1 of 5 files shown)


Daguerreo:
It could be interesting adding it in the calculator example. Should be quite simple, connecting the `modified` signal to the `QWidget::setWindowModified(bool)`.

`connect(scene, &BasicGraphicsScene::modified, this, [this](){ setWindowsModified(true); };`

QianMoth:
@Daguerreo I have made changes to the calculator example.
In order to be able to remove the flag when saving and keep it when saving aborts. I had to make changes to the `save()` and `load()` functions
image

PR #371, opened 2023-02-14
Add option to make the drop shadow effect optional (fix performance)

arabine:
This change add an option to the GraphicsScene class to enable or disable the drop shadow effect for all the items.
This fixes performance issue when zooming, at least on Ubuntu 22.10.

Changes:

- getter/setter to enable globally the drop shadow effect on the GraphicsScene class
- Added margins ratio member and setter to nicely fit the bounding rect to the widget node boundaries (kept the default to 0.2 for compatibility)


arabine:
New commit: added margins ratio modier to allow set it to zero (usefull when the drop shadow effect is disabled)

Issue #397, opened 2023-09-19
Qt dependency problem


DarellLdark:
If we have Qt5 and Qt6 versions installed at the same time on build machine, the library will be build with Qt5
It may be a problem if your project forced Qt6 version, since ABI will be different

CMakeLists.txt contains the following:
```
find_package(QT NAMES Qt6 Qt5 REQUIRED COMPONENTS Widgets)
find_package(Qt${QT_VERSION_MAJOR} REQUIRED COMPONENTS Core Widgets Gui OpenGL)
message(STATUS "QT_VERSION: ${QT_VERSION}, QT_DIR: ${QT_DIR}")
```

This will work as described:
1. Since we have two entries to find (Qt6 and Qt5), there will be two search cycles
2. First iteration will set `QT_VERSION_MAJOR` to 6
3. Second iteration will set `QT_VERSION_MAJOR` to 5
As result, the library will always be build with Qt5 instead of Qt6

Solution:
Maybe we can add a flag, which will allow users to use selected Qt version? For example, it may be like this:
`option(USE_QT6 "Build with Qt6 (Enabled by default)" ON)`

So as library user I can just build with Qt5 if I want to and use default build with Qt6

P.S. Submitted PR, let me know what do you think about it

PR #398, opened 2023-09-19
Fix qt5/qt6 dependency problem

DarellLdark:
#397 

This seems to be intuitive option, plus adds some control for library users to use the versions they want

Added flag for switching Qt versions + readme update on this info


paceholder:
Cleaned up and merged in https://github.com/paceholder/nodeeditor/pull/415

PR #401, opened 2023-09-28
Change some `unsigned int` variables to `NodeId`.

zmoth:
Fix the `getNodeId()` function return value to `NodeId`

`toInteger()` is used because you want to be compatible with `uint64_t`


paceholder:
1. `toInteger` does not return an unsigned number. It has nothing to do with the compatibility with `uint64_t`.
2. Qt5 which is still supported by this library does not have the function `toInteger`

paceholder:
I merged some changes from this PR

PR #402, opened 2023-10-10
fix: name of paste selection GraphicsView action

### LINKED ISSUE #372: The text is wrong, the copy and Paste are wrong
I found a text error in one place:
https://github.com/paceholder/nodeeditor/blob/master/src/GraphicsView.cpp
line,132,_pasteAction = new QAction(QStringLiteral("Copy Selection"), this);

The correct way to write it is:
"Copy Selection"  --> "Paste Selection"

craftablescience:
The paste selection action was incorrectly named "Copy Selection"

This fixes #372


nolankramer (approved):
This is a good fix

PR #403, opened 2023-11-28
Use nodeId style instead the generic StyleCollection style

Dalmais:
If we want to customize a nodeId style then we should return the data of the nodeId instead of the generic StyleCollection style. By default the nodeId style use the generic StyleCollection so there is no change for user that don't tweak the style.


mikee1976:
Yes, it took me some time to identify why it is not working in my case with custom style. So please add this to the main branch.

nolankramer (approved):
This is a good fix

PR #421, opened 2024-05-13
[Bug Fix] Update embedded QWidget position after node update

nolankramer:
Currently, if a node's inputs or outputs are updated, the embedded QWidget's position will remain unchanged - possibly occluding the new ports' captions.

This code fixes that issue, and moves the embedded QWidget with the new geometry after a node has changed.

## Before
image

## After
image


paceholder:
Thanks for this fix

PR #427, opened 2024-06-13
Optimizing performance | Add shadow enabler

Grozamir:
In my app if count nodes more 30+, so then starting freezing and downing performance. The culprit of all this's shadows. I added enabler shadow. And now in my project disabled shadows and i can spawn very more nodes

There was a problem:
image

### DIFF
--- a/docs/features.rst
+++ b/docs/features.rst
@@ -196,6 +196,7 @@ the moment of writing this documentation.
       "GradientColor2": [64, 64, 64],
       "GradientColor3": [58, 58, 58],
       "ShadowColor": [20, 20, 20],
+      "ShadowEnabled": false,
       "FontColor" : "white",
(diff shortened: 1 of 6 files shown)


Daguerreo:
Is it everything really necessary? It should be enough to check if the "ShadowColor" key is valid/populated and only in that case instance the shadow effect.

Grozamir:
It seems to me that it was natural to create a shadow switch.

Daguerreo:
> It seems to me that it was natural to create a shadow switch.

and what if the switch is true and the shadow color not setted?

nolankramer:
This is a good idea for optimizing based on user's preference between appearance/drawing performance - however it may also warrant implementing shadow drawing ourselves - since it's likely we can make assumptions that Qt isn't aware of.

PR #439, opened 2024-09-16
Fix header inclusion paths

nolankramer:
Some of the headers are under the `src/` directory.

This leads to issues when including them using the redirect headers under `include/QtNodes` - such as `DefaultNodePainter`.

As an example, `#include <QtNodes/DefaultNodePainter>` currently contains this code:
```
#include "internal/DefaultNodePainter.hpp"
```

Which obviously points to `internal/DefaultNodePainter.hpp`. However, for whatever reason, that file is currently under `src/`, so the include path is broken.

This PR fixes all cases that are like this.


paceholder:
Totally valid. Thanks a lot, merged

Issue #442, opened 2024-09-19
crashed when running the calculator


CuberFan:
in the calculator, there is a crash when I connect the node "NumberSoure" and node “Addition”

QT  6.7.2
Windows 10
MSVC 2019

![20240919154147](https://github.com/user-attachments/assets/da0af931-77e7-434c-ab74-e744918393da)

chateauferret:
That looks pretty similar to #441 which I raised yesterday. [https://github.com/paceholder/nodeeditor/issues/441](https://github.com/paceholder/nodeeditor/issues/441
)

CuberFan:
> That looks pretty similar to #441 which I raised yesterday. #441

thanks!

paceholder:
I think this is fixed now. Please reopen if the problem persists

Issue #441, opened 2024-09-19
Null pointer on connecting nodes in certain examples


chateauferret:
In some of the examples, e.g. the calculator, there is a crash because of a null ProxyWidget pointer in places like this in NodeGraphicsObject.cpp line 81:

void NodeGraphicsObject::updateQWidgetEmbedPos()
{
    _proxyWidget->setPos(nodeScene()->nodeGeometry().widgetPosition(_nodeId));
}

I might try and fix that by simply testing the pointer for NULL before it is dereferenced, but thought it wise to consult you first.

chateauferret:
That's on Qt 6.7.2 in Microsoft Windows (and seen on two different computers btw).

chateauferret:
OK, so it seems to behave if I just make sure the `ProxyWidget `pointer isn't `NULL `before I try to dereference it. There are a few other places where I think this is a risk, because some of the other examples are crashing at similar but different places. So I'll go through and look for them then see if I can put down a pull request. 

Root cause seems to be line 91, in the method `embedQWidget ` where `_graphModel.nodeData` must be returning `NULL`.

Think we can fix this without too much bother then.

chateauferret:
Slightly better, let's suggest this `embedQWidget `- always assign a value to the pointer.

```
void NodeGraphicsObject::embedQWidget()
{
    AbstractNodeGeometry &geometry = nodeScene()->nodeGeometry();
    geometry.recomputeSize(_nodeId);
    _proxyWidget = new QGraphicsProxyWidget(this);
    if (auto w = _graphModel.nodeData(_nodeId, NodeRole::Widget).value<QWidget *>()) {
        _proxyWidget->setWidget(w);
        ....
```

Daguerreo:
> Slightly better, let's suggest this `embedQWidget `- always assign a value to the pointer.
> 
> ```
> void NodeGraphicsObject::embedQWidget()
> {
>     AbstractNodeGeometry &geometry = nodeScene()->nodeGeometry();
>     geometry.recomputeSize(_nodeId);
>     _proxyWidget = new QGraphicsProxyWidget(this);
>     if (auto w = _graphModel.nodeData(_nodeId, NodeRole::Widget).value<QWidget *>()) {
>         _proxyWidget->setWidget(w);
>         ....
> ```

If embedQWidget() is called multiple time you continue to allocate a new QGraphicsProxyWidget. It should be included a guard at least to destroy the previous pointer.

[...]

Issue #417, opened 2024-04-06
connect exception


jorketcode:
Connecting the output of the following nodes to the preceding nodes will throw an exception

chateauferret:
See https://github.com/paceholder/nodeeditor/issues/441.

Llcoolsouder:
Duplicate of #198. Feel free to move the discussion over there/subscribe to that ticket for updates.

PR #443, opened 2024-09-19
Fixed NULL pointer exception on connecting nodes (https://github.com/…

chateauferret:
…paceholder/nodeeditor/issues/441)

Moved the creation of the ProxyWidget in NodeGraphicsObject::embedQWidget() (line 84 ff.) out of the if { } block to ensure that it wouldn't be NULL when someone tried to dereference it. See issue #441. May be relevant also for #417 and #442.


chateauferret:
Added null check (in my fork) around the line in question to avoid multiple allocations - see comment on #441.

nolankramer (approved):

nolankramer (approved):

paceholder:
Fixed in master by checking for `nullptr` in ` void NodeGraphicsObject::updateQWidgetEmbedPos()`

PR #451, opened 2024-11-04
DataFlowGraphModel.hpp , DataFlowGraphModel.cpp , UndoCommands.cpp

qq978358810:
1:When you add a connection, you make security judgments about the data


qq978358810:
1: Added topology connection checks to prevent cyclic connections.

2: When calling addConnection, data consistency must be verified, and it should be checked if a connection already exists (since another workflow might impact synchronization with the current workflow).

Llcoolsouder:
@qq978358810 Can you confirm that this is different from #484 ?

PR #465, opened 2025-04-28
useful if we want to inherit BasicGraphicsScene with similarities to DataFlowGraphicsScene

clythersHackers:
See title

### DIFF
--- a/include/QtNodes/UndoCommands
+++ b/include/QtNodes/UndoCommands
@@ -1,0 +1,1 @@
+#include "internal/UndoCommands.hpp"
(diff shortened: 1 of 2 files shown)


Llcoolsouder:
@clythersHackers This seems harmless enough. A little more explanation as to what this gives you in terms of BasicSceneGraph would be super helpful though.

clythersHackers:
> @clythersHackers This seems harmless enough. A little more explanation as to what this gives you in terms of BasicSceneGraph would be super helpful though.

It's a while ago so... reload context..., IIRC there's a relatively high-level DataFlowGraphicsScene with predefined behaviours which I wanted to customise myself, so I wanted to inherit directly from BasicGraphicsScene, for example drag & drop behaviours, right mouse click, etc. But trying to do that I found BasicGraphicsScene interfaces were private.

Issue #226, opened 2019-04-18
Signals & working flow


starkimax:
Hey guys !

Thank you for this amazing project ! 
I had a question, is there any way I can update the graphic of the validation message without moving the node or creating a new connecting to this node ? 

I'm asking this because I have nodes that can be updated without any new connection, they are updated when I edit my node widget.

![image](https://user-images.githubusercontent.com/42606250/56347575-db2aa480-61c4-11e9-9e18-e157dc775dc0.png)

It would be great if you could create a signal that would call your "drawValidationRect" function !
If your ok with it I can try to implement it on master branch.

Thx in advance ! Great job

paceholder:
Actually, I wanted to remove this graphical message because it looks ugly.
The better way to emphasize an input error would be possibly to highlight the input point with red..

Daguerreo:
> Actually, I wanted to remove this graphical message because it looks ugly.
> The better way to emphasize an input error would be possibly to highlight the input point with red..

Highlighting the missing input would be nice for sure, but I find the validation message can be still useful though. Or at least a generic runtime message :D

MaxenceEpitech:
Well, the graphical message is maybe ugly, but I in my case It's very usefull to help the user know what he did wrong. I'll try to highlight the input as red, thx.
In my opinion, removing this graphical message would be a bad idea, it would remove possibilities for the user.

tatatupi:
I would like to revisit this discussion. Considering @paceholder opinion on the graphical message being ugly, I implemented a tooltip behavior in PR #475 with a color scheme.

In my opinion, the color feedback should concern the entire node body, rather than the connection or port. I believe that an invalid state is a reflection of a combination of inputs and node configuration. An input is never wrong by itself. For the calculator example, dividing by 0 produces an invalid state. But the number 0 is not problematic itself.

In some applications, the invalid state occurs because the two inputs are not valid in relation to each other, rather than a single input being the cause of the error. 

![Image](https://github.com/user-attachments/assets/43a51afc-dc24-4680-a32b-31a1ffab7959)

Issue #75, opened 2017-03-22
Graphics glitch when validation state changes


russelltg:
![bug](https://cloud.githubusercontent.com/assets/7105082/24225094/3a8e85cc-0f24-11e7-8d7f-bca433f83a19.gif)

Pretty small bug, but it looks a little funky. 

Easy workaround of just dragging on one of the nodes.

paceholder:
I don't like this jumpy nodes and large red notification area. It breaks the elegance and simplicity of the framework. I will find the time on the weekend and re-do it properly (with tool-tips and highlighted connection sockets).

russelltg:
I agree, the jumpy nodes can be really annoying (I had several cases where I found a node behind another node because the validation state changed). plus there are definitely some graphical bugs.

russelltg:
For me it'd be nice to be able to click on the error tooltip after it comes up and click on links in the error message.

That'd be useful because I'm looking to add "quickfixes" to errors, so people can click a button to fix it immediately for simple errors.

PR #475, opened 2025-06-22
Adds validation state feedback to nodes

tatatupi:
The idea is to have a `NodeValidationState` member object in the `NodeDelegateModel` class, that defaults to a valid and empty state message. The API allows for updates to be made accordingly, and the UI will respond automatically.

UI will show a node red (error) or yellow (warning). A tooltip is added with the state message. And a small icon for information is displayed as a badge, to help the user access the tooltip. I added a screenshot of the calculator example to help discussions. in the example, the tooltip for the Warning state is displayed.

Key Changes

- Introduces `NodeValidationState` struct class.
- Getter and setter in NodeDelegateModel for validation.
- `DataFlowGraphModel` handles `NodeRole::ValidationState` updates.
- Painting logic reflects invalid states in `DefaultNodePainter`.
- Tooltip shows validation message in `NodeGraphicsObject`.
- Calculator example demonstrates usage.

image

Solves #226 
Solves #75 

P.S.: some changes happened automatically when saving files, due to autoformatting using nodeeditor clang format style.


tatatupi:
I also changed the linux-gcc qt version to 6, so that QtGui/QUndoCommand include could work

Llcoolsouder (approved):
Everything appears to be building and working for me.

Just that one note above about the QtGui/X.h includes. Doesn't look like you've added any of those includes here, so this should build with or without the version change. If anything I'd recommend adding both major Qt versions back to the build matrix, but not strictly necessary.

PR #484, opened 2025-08-01
Control loops before connecting depending on AbstractGraphModel::loopsEnabled() 

paceholder:
First commits brings test-classes implementation to the definition of these classes in the header for simplicity.
Second commit:
  - extends unit-tests that tries to make a loop
  - implemetns depth-first search to detect a recursion in DataFlowGraphModel::connectionPossible

### DIFF
--- a/test/src/TestDataFlowGraphModel.cpp
+++ b/test/src/TestDataFlowGraphModel.cpp
@@ -113,6 +113,29 @@ TEST_CASE("DataFlowGraphModel connections", "[dataflow]")
         ConnectionId invalidIn{node1, 0, node2, 2};
         CHECK_FALSE(model.connectionPossible(invalidIn));
     }
+
+    SECTION("Loop connection between three nodes")
+    {
+      NodeId node3 = model.addNode("TestNode");
+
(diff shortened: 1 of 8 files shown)


test/src/TestDataFlowGraphModel.cpp
```diff
@@ -113,6 +113,29 @@ TEST_CASE("DataFlowGraphModel connections", "[dataflow]")
...
+
+      CHECK(model.connectionPossible(connId23));
+      model.addConnection(connId23);
+      CHECK(model.connectionExists(connId23));
+
+      ConnectionId connId31{node3, 0, node1, 0};
+
+      CHECK_FALSE(model.connectionPossible(connId31));
+      model.addConnection(connId31);
+      CHECK(model.connectionExists(connId31));
```
Llcoolsouder, line 137: Shouldn't this be a `CHECK_FALSE`?
```suggestion
      CHECK_FALSE(model.connectionExists(connId31));
```

Llcoolsouder (approved):
Code looks good. Tests run successfully. Calculator example is rejecting feedback loops.

Looks good to me!

PR #486, opened 2025-08-05
Adds find_package(Catch2) to test_nodes CMakeLists

tatatupi:
## Type of change
- [x] Bug fix
- [x] Documentation/refactoring

## Description
In case you have Catch2 installed and the correct system variable for `Catch2_DIR`, the `test_nodes` lib won't find the Catch2 package without the explicit directive to see it. 
This PR includes `find_package(Catch2)` to test_nodes CMakeLists.txt to fix that.
And it also adds instructions in README.txt, to create the Catch2_DIR system variable.

## Testing
- Qt version tested: 
- [x] Existing tests still pass

## Breaking changes?
No

## Related issue
None.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,7 +74,7 @@ Dependencies
 
 * Qt >5.15
 * CMake 3.8
-* Catch2
+* Catch2 (add `Catch2_DIR` to the system variable, pointing to the folder containing `Catch2Config.cmake`)
 
 
 Current State (v3)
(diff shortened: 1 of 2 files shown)


tatatupi:
needs more elaboration

paceholder:
`CMakeListst.txt` - line 50
`add_subdirectory(external)` -- there we run the script `external/CMakeLists.txt` which already searches for catch2.

This is maybe not very clear from the name (external), maybe we should move the fings outside the external directory

tatatupi:
The problem is that I have Catch2 3.6.0 installed, and the configuration won't work due to the different version. I propose we revert the `find_package(Catch2 2.13.7 QUIET)` with an explicit version.

And when the system package is found, there should be a `find_package(Catch2)` inside the test CMakeLists.

PR #495, opened 2025-10-30
BUGFIX: GraphicsView::setScene(nullptr)

cxxamz:
## Type of change
- [x] Bug fix

## Description
GraphicsView::setScene support nullptr

## Testing
- Qt version tested: 
- [x] Existing tests still pass

## Breaking changes?
None.

## Related issue
None.

---
*Please fill out the sections above to help reviewers understand your changes.*

### DIFF
--- a/src/GraphicsView.cpp
+++ b/src/GraphicsView.cpp
@@ -75,116 +75,93 @@ QAction *GraphicsView::deleteSelectionAction() const
 void GraphicsView::setScene(BasicGraphicsScene *scene)
 {
     QGraphicsView::setScene(scene);
+    if (!scene)
+    {
+        // Clear actions.
+        delete _clearSelectionAction;
+        delete _deleteSelectionAction;
+        delete _duplicateSelectionAction;
+        delete _copySelectionAction;
+        delete _pasteAction;
+        _clearSelectionAction = nullptr;
+        _deleteSelectionAction = nullptr;
+        _duplicateSelectionAction = nullptr;
+        _copySelectionAction = nullptr;
+        _pasteAction = nullptr;
+        return;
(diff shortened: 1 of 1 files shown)


src/GraphicsView.cpp
```diff
@@ -274,27 +315,29 @@ void GraphicsView::setupScale(double scale)
 
 void GraphicsView::onDeleteSelectedObjects()
 {
-    nodeScene()->undoStack().push(new DeleteCommand(nodeScene()));
+    if (auto scene = nodeScene()) scene->undoStack().push(new DeleteCommand(scene));
```
paceholder: Is there any valid configuration when `scene != nodeScene`?
Shouldn't we better assert this condition?

src/GraphicsView.cpp
```diff
@@ -274,27 +315,29 @@ void GraphicsView::setupScale(double scale)
...
 }
 
 void GraphicsView::onDuplicateSelectedObjects()
 {
     qDebug() << "ON DUPLICATE";
```
paceholder: TODO: maybe remove debug output or write a better message without capslock

src/GraphicsView.cpp
cxxamz: Is this acceptable?

```C++
void GraphicsView::onDeleteSelectedObjects()
{
    if (!nodeScene()) return; 

    nodeScene()->undoStack().push(new DeleteCommand(nodeScene()));
}
```

src/GraphicsView.cpp
cxxamz: OK, I removed this message, even though it wasn't written by me.

